PR #9, opened 2016-07-13
Organize components

cheshire137:
@probablycorey I made directories for each component and put them in a components/ directory. For each component directory, I added a package.json that lets us require the directory as a whole instead of `Filter/Filter`. We don't have to do it this way if you don't like it.


components/Main/Main.jsx
```diff
@@ -1,8 +1,8 @@
 const React = require('react')
-const Filter = require('./filter')
-const TaskList = require('./task-list')
+const Filter = require('../Filter')
+const TaskList = require('../TaskList')
 
-class App extends React.Component {
+class Main extends React.Component {
```
probablycorey, line 5: I avoid using Main in electron apps because of the main.js file. It always confused me when there were multiple mains.

probablycorey:
I can see how the per-component dirs make sense when each has a css file, but it seems a little premature right now since we don't have any css files!

But I'm assuming we are going to add css files very soon, so I'm not completely against it.

cheshire137:
> I avoid using Main in electron apps because of the main.js file.

I could rename the Main component. Maybe Home, Landing, SignedIn? It's going to be the "main" page when the user is signed in.

> I'm assuming we are going to add css files very soon

Getting our CSS pipeline set up is my next goal! 👍

PR #10, opened 2016-07-14
Faster tests

probablycorey:
This PR does two things.
1. Renames Main to App (based on the comment in https://github.com/cheshire137/gh-notifications-snoozer/pull/9#issuecomment-232754084). Since all electron apps are single page apps, I think is a good name.
2. Uses electron-mocha instead of spectron. This will make are tests much faster. And if they are a lot faster, I am more likely to use them!

**spectron**

``` sh
> mocha



  application launch
    ✓ shows an initial window
    1) renders sarah


  1 passing (3s)
  1 failing

  1) application launch renders sarah:
     NoSuchElement: An element could not be located on the page using the given search parameters.
      at elements("p") - getText.js:51:17
      at getText("p") - main-test.js:29:28



npm ERR! Test failed.  See above for more details.
npm test  1.43s user 0.35s system 41% cpu 4.320 total
```

**electron-mocha**

``` sh
> electron-mocha --recursive --compilers js:babel-register test



  App
    ✓ renders


  1 passing (70ms)

npm test  1.77s user 0.27s system 105% cpu 1.941 total
```

4.3s vs 1.9s. It's not much now, but spectron opens a new electron window for every test, and that would become super slow.

cc @cheshire137


components/App/App.jsx
```diff
@@ -2,7 +2,7 @@ const React = require('react')
 const Filter = require('../Filter')
 const TaskList = require('../TaskList')
 
-class Main extends React.Component {
+class App extends React.Component {
```
cheshire137, line 5: I had originally gone away from app.jsx because I find it a little confusing. There's main.js, snoozer.js, and app.jsx. All are kind of entry points for the app. This component in particular is the main page of the app, but only for signed-in users.

I'm okay sticking with App for now, but I think this could be a little clearer. That could wait till we have the concept of logging in with your GitHub account, though, and have to distinguish between the signed-in versus signed-out landing page.

test/components/app-test.js
```diff
@@ -0,0 +1,19 @@
+const assert = require('assert')
+
+const App = require('../../components/app/app')
```
cheshire137: I think your case insensitive file system is showing. :D 1) This should be `../../components/App/App` but 2) it could really be `../../components/App` because of the package.json in the App/ directory.

test/components/app-test.js
```diff
@@ -0,0 +1,19 @@
...
+const TestUtils = require('react-addons-test-utils')
+
+describe('App', function() {
+  before(function() {
+    global.document = jsdom.jsdom('<!doctype html><html><body></body></html>')
+    global.window = global.document.defaultView
+  })
+
+  it('renders', function() {
+    const appComponent = TestUtils.renderIntoDocument(<App/>)
```
cheshire137, line 16: Does `renderIntoDocument` just shove the given React node into the `<body>`?

test/components/app-test.js
```diff
@@ -0,0 +1,19 @@
...
+const App = require('../../components/app/app')
+const jsdom = require('jsdom')
+const React = require('react')
+const ReactDOM = require('react-dom')
+const TestUtils = require('react-addons-test-utils')
+
+describe('App', function() {
+  before(function() {
+    global.document = jsdom.jsdom('<!doctype html><html><body></body></html>')
+    global.window = global.document.defaultView
```
cheshire137, line 12: Ooh, is `global` some new JavaScript thing?

cheshire137:
I approve of this test _sweet._ 😆

components/App/App.jsx
probablycorey, line 5: > There's main.js, snoozer.js, and app.jsx

I agree, it is confusing. I got rid of `snoozer.js` but the logic is now in `index.html` so it still is confusing. I was thinking that everything would start with the App component, even signed-out users. If the user is logged out it will render the login page, if they are logged in it will render the logged in page.

But I'm totally willing to change this if it ends up not feeling right.

test/components/app-test.js
probablycorey, line 16: Based on https://facebook.github.io/react/docs/test-utils.html#renderintodocument, I think it creates the component in isolation.

test/components/app-test.js
probablycorey, line 12: It's a node thing https://nodejs.org/api/globals.html#globals_global. It's basically the `window` var of node.

test/components/app-test.js
probablycorey: I had to change the package.json file in `App`, but it's working now 👍

PR #12, opened 2016-07-14
Add .env requirement

### LINKED ISSUE #3: Allow signing in with GitHub
To get a list of issues and PRs that suit a user's rules, we need to authenticate with GitHub as that user.

probablycorey:
This PR does two things.
1. Makes sure you have a .env file with a token in it. This is for dev purposes only, so we don't have to spend time working on the oauth flow.
2. Remove the `snoozer.js` file and move its logic into the `index.html` file. I did this because the snoozer logic is so minimal.

Closes #3

cc @cheshire137

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 npm-debug.log
 node_modules
 .DS_Store
+.env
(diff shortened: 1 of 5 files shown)


cheshire137:
👍

PR #13, opened 2016-07-14
Start on a CSS pipeline with LESS

cheshire137:
This branch adds:
- Travis CI setup for tests and style checker
- LESS for styling
- Fixes style checker complaints
- Organizes code into src/ directory

Hmm, so the app starts, but the styling doesn’t seem to be applied. Any thoughts on this, @probablycorey?


.compilerc
```diff
@@ -0,0 +1,9 @@
+{
```
probablycorey, line 1: Is this a `electron-compile` thing? It's pretty nice.

src/index.html
```diff
@@ -2,7 +2,8 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>Hello World!</title>
+    <title>GitHub Notifications Snoozer</title>
+    <link rel="stylesheet" type="text/less" href="app.less">
```
probablycorey: Maybe this should be `src/app.less` or `./src/app.less`?

src/main.js
```diff
@@ -15,7 +15,7 @@ function createWindow () {
   // and load the index.html of the app.
   mainWindow.loadURL(`file://${__dirname}/index.html`)
 
-  // mainWindow.webContents.openDevTools()
+  mainWindow.webContents.openDevTools()
```
probablycorey: **protip:** You can press `cmd+opt+i` to open up the console

.compilerc
cheshire137, line 1: Yep! See https://github.com/electron/electron-compile#how-do-i-set-up-babel--less--whatever-the-way-i-want

src/index.html
cheshire137: Hmm, just tried both, no luck. Still a white background everywhere!

src/main.js
cheshire137: Ah yeah, I was doing that then got lazy. I'll comment this back out, didn't intend to check it in!

cheshire137:
I spoke with @zeke about this yesterday, he thought he might be able to help!

zeke:
👋 I haven't figured out the less issue yet, but I did notice that `electron` is in the dependencies, which could be screwing things up. That package name is a different project. Fortunately, that user has graciously handed over the `electron` npm name to us, but we haven't yet published something like `electron-prebuilt` in its place.

cheshire137:
Hm, I got this branch updated with master just now and `electron` is no longer a dependency in package.json.

zeke:
The Less configuration seems right, but I'm not seeing the styles.

I tried debugging with `DEBUG=*,-babel npm start`, but didn't see any issues in the logs.

@paulcbetts do you see any obvious signs of misconfiguration here?

anaisbetts:
I'll have a look in a hot second

src/components/App/App.less
```diff
@@ -0,0 +1,5 @@
+@import "../Filter/Filter.less";
```
anaisbetts, line 1: It's probably easier to just include this inline in the JSX file:

``` html
<style type="text/less">
@import '../filter/filter.less';
...
</style>
```

anaisbetts:
@cheshire137 Ok, tracked this down, it turns out I am not good at computer

anaisbetts:
Ok! Blow away `node_modules` and reinstall so that you get `electron-compilers@3.1.6` and this should just work. You should also go into $TMPDIR and remove any folder that starts with `compileCache`

cheshire137:
Thank you, it works!

screen shot 2016-07-20 at 1 22 50 pm

I checked out [your commits](https://github.com/electron/electron-compilers/compare/8b77d4cb5ed5c3843f6efcc3fb158f442e053f92...3d9498f2dbd7da1e7c64605d0235ce3101f572ae)--thanks for helping!

zeke:
👏

package.json
```diff
@@ -12,13 +12,19 @@
     "electron-prebuilt-compile": "^1.2.5",
     "eslint": "^3.0.1",
     "eslint-config-airbnb": "^9.0.1",
+    "eslint-plugin-import": "^1.11.0",
+    "eslint-plugin-jsx-a11y": "^2.0.1",
+    "eslint-plugin-react": "^5.2.2",
```
probablycorey, line 17: Do we need these three packages? I think they are included with the `eslint-config-airbnb` package's dependencies.

cheshire137, line 17: We do! Check the full commit messages, tests failed on Travis till these were added.

anaisbetts, line 17: Is Travis running old-ass npm 2 or something?

cheshire137, line 17: Haha I think my Travis config file specifies node 6. Check a build, it spits out node, npm, and nvm version numbers.

PR #14, opened 2016-07-16
Use redux to list tasks

### LINKED ISSUE #4: Allow viewing issues and PRs that match a rule
I should be able to view a list of issues and pull requests that match each rule -- see #2. Clicking an issue or PR should take me to its page on github.com.

probablycorey:
This is a SUPER WIP for adding redux. I'm just opening the PR now to give visibility to me working on it. When I'm ready for review I'll write a new comment and update this body.

Closes #4


components/TaskList/TaskList.jsx
```diff
@@ -1,20 +1,42 @@
...
-          <button>archive</button>
-        </nav>
-        <ol>
-          <li>LOOK AT THIS ISSUE</li>
-        </ol>
-      </div>
-    )
+const TaskList = ({tasks}) => {
+  const taskElements = tasks
+    .filter((task) => {
```
cheshire137: Oh no! The period-on-the-next-line strikes again!

J/k, BTW, it's cool.

components/TaskList/TaskList.jsx
```diff
@@ -1,20 +1,42 @@
...
+      <ol>
+        {taskElements}
+      </ol>
+    </div>
+  )
+}
+
+const mapStateToProps = (tasks) => {
+  return {
+    tasks: tasks
```
cheshire137: Our linter might fuss at you for this. You should be able to do `{ tasks }`.

test/components/TaskList-test.js
```diff
@@ -0,0 +1,47 @@
...
+    global.document = jsdom.jsdom('<!doctype html><html><body></body></html>')
+    global.window = global.document.defaultView
+
+    const appComponent = TestUtils.renderIntoDocument(<TaskList store={store}/>)
+    element = ReactDOM.findDOMNode(appComponent)
+
+    assert(element)
+  })
+
+  it('do not show tasks that are ignored, archived or snoozed', function() {
```
cheshire137: It do not. Maybe `does`? :D

components/TaskList/TaskList.jsx
probablycorey: I'm following the AirBnb style guide!

`Use indentation when making long method chains. Use a leading dot, which emphasizes that the line is a method call, not a new statement.`

components/TaskList/TaskList.jsx
probablycorey: The linter is very unhappy with me right now.

src/components/Task/Task.jsx
```diff
@@ -0,0 +1,41 @@
+const React = require('react')
+const { connect } = require('react-redux')
+
+class Task extends React.Component {
+  onChange(event) {
+    const { task } = this.props
+    const type = event.target.checked ? 'TASKS_SELECT' : 'TASKS_UNSELECT'
```
cheshire137: Maybe `DESELECT`?

src/components/Task/Task.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+      <li>
+        <input type="checkbox" onChange={event => this.onChange(event)} />
+        <span>{task.title}</span>
+      </li>
+    )
+  }
+}
+module.exports = connect()(Task)
```
cheshire137: Ooh strange.

src/components/Task/Task.jsx
```diff
@@ -0,0 +1,41 @@
+const React = require('react')
```
cheshire137: Instead of adding this new class/file, you could overwrite TaskListItem that got added in https://github.com/cheshire137/gh-notifications-snoozer/pull/16.

src/components/TaskList/TaskList.jsx
```diff
@@ -1,20 +1,31 @@
...
         </ol>
       </div>
     )
   }
 }
 
-module.exports = TaskList
+const mapStateToProps = tasks => ({ tasks })
+
+module.exports = connect(mapStateToProps)(TaskList)
```
cheshire137, line 43: Can you explain what's happening with lines 29-31?

src/reducers/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+      return tasks.map(task => {
+        if (task.id === action.task.id) {
+          return Object.assign({}, task, { isSelected: true })
+        }
+        return task
+      })
+    case 'TASKS_UNSELECT':
+      return tasks.map(task => {
+        if (task.id === action.task.id) {
+          return Object.assign({}, task, { isSelected: false })
```
cheshire137, line 26: What's this doing?

test/components/TaskList-test.js
```diff
@@ -0,0 +1,61 @@
...
+    element = ReactDOM.findDOMNode(appComponent)
+
+    assert(element)
+  })
+
+  it('does not show tasks that are ignored, archived or snoozed', function() {
+    assert.equal(2, element.querySelectorAll("li").length)
+  })
+
+  it('unarchives a task if the takes has updates', function() {
```
cheshire137:
``` diff
- takes
+ task
```

test/components/TaskList-test.js
```diff
@@ -0,0 +1,61 @@
...
+    const tasks = defaultTasks.map(task => {
+      task.updateAt = new Date()
```
cheshire137: Is `updateAt` the field name we'll get back from the GitHub API? I think it uses snake_case, at least results from the /notifications endpoint.

test/components/TaskList-test.js
```diff
@@ -0,0 +1,61 @@
...
+      return task
+    })
+
+    store.dispatch({type: 'TASKS_UPDATE', tasks: tasks})
+    assert.equal(3, element.querySelectorAll("li").length)
+  })
+
+  context('when the snooze button is pressed', function() {
+    it('snoozes the selected tasks', function() {
+      const snoozeButton = element.querySelector('button#snooze')
```
cheshire137: Could you use `element.getElementById('snooze')`?

src/reducers/Tasks-test.js
```diff
@@ -1,21 +0,0 @@
-const assert = require('assert')
```
cheshire137: Re: the commit message

dog-no-idea

probablycorey:
I think getting this PR in and getting the notifications changes in there might become too big. So I'm going to add the actual notification data in a new PR and update this to use dummy data. It will be easier to explain that way.

test/components/TaskList-test.js
probablycorey: I got rid of all of the component tests because I'm not sure of the best way to test them right now.

I think I'll try and use https://github.com/airbnb/enzyme to test components in a follow up PR. It was the recommended framework to use in the React Docs https://facebook.github.io/react/docs/test-utils.html.

test/components/TaskList-test.js
probablycorey: For now it is using the dummy data found in https://github.com/cheshire137/gh-notifications-snoozer/pull/14/files#diff-053da0fe3233c4c7a2ad76216a682b65R2.

I want to get the real endpoints hooked up in a new PR. When we do I'd like to force the GitHub API result objects into a camelcase format so it looks like regualr javascript.

src/reducers/reducer.js
probablycorey, line 26: Redux has three main rules http://redux.js.org/docs/introduction/ThreePrinciples.html

Because state can't be mutated, this creates a new object based on the original task and adds the `isSelected` field. It is similar to underscore's `extend` function.

src/components/TaskList/TaskList.jsx
probablycorey, line 43: The best way is to read http://redux.js.org/index.html :wink: but I'll try a tl;dr.

With Redux, an app has one object that holds every bit of state. You can pass this state from component to component via props, but that gets a little cumbersome. So in redux people use `connect`. It gives a component access to the global state by adding it to the components props. If you don't want to give the component access to the entire state, you can reduce what it has access to using `mapStateToProps`.

After explaining that I realize I don't need to use mapStateToProps here, so I'm going to remove it.

probablycorey, line 43: > After explaining that I realize I don't need to use mapStateToProps here, so I'm going to remove it.

I'm actually going to leave this as is for now since I can't figure out how to get the state out of the props without using `mapStateToProps`

src/components/Task/Task.jsx
probablycorey: I don't really like UNSELECT or DESELECT, but I'll go with DESELECT for now

probablycorey:
Ok, this is now ready. It is mostly just boilerplate for future usage, but I would rather keep the PR simple and put more logic into followup PRs. If this looks good to you, we can follow it up with:
1. A PR that uses the GitHub API to populate the tasks
2. A PR that uses enzyme to test components
3. A PR that uses [actions](http://redux.js.org/docs/basics/Actions.html) instead of raw calls to `dispatch`
4. A PR that adds tests the actions and reducer

test/components/TaskList-test.js
```diff
@@ -0,0 +1,20 @@
...
+
+  context('when the snooze button is clicked', () => {
+    it('hides selected tasks')
+    it('updates the selected task\'s `snooze_until` field')
+    it('shows the tasks again, starting at midnight of the next day')
+  })
+
+  context('when the archive button is clicked', () => {
+    it('hides selected tasks')
+    it('updates the selected task\'s `archived_at` field')
```
cheshire137, line 12: Fear not the double quote!

cheshire137:
Here we go!

probablycorey:
💥

PR #16, opened 2016-07-20
List notifications from GitHub

cheshire137:
This adds a GitHub model for us to start hitting the GitHub API. It also lists all the user's notifications in a very simple format for now.

/cc @probablycorey

### DIFF
--- a/src/components/TaskList/TaskList.jsx
+++ b/src/components/TaskList/TaskList.jsx
@@ -32,9 +32,9 @@ class TaskList extends React.Component {
           <button type="button" className="control">archive</button>
         </nav>
         <ol className="issues-list">
-          {this.state.notifications.map(notification => {
-            return <TaskListItem {...notification} key={notification.id} />
-          })}
+          {this.state.notifications.map(notification =>
+            <TaskListItem {...notification} key={notification.id} />
+          )}
         </ol>
       </div>
     )
--- a/src/components/TaskList/TaskList.jsx
+++ b/src/components/TaskList/TaskList.jsx
@@ -1,5 +1,6 @@
 const React = require('react')
 const GitHub = require('../../models/github')
+const TaskListItem = require('../TaskListItem')
 
 class TaskList extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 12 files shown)


src/components/TaskList/TaskList.jsx
```diff
@@ -1,6 +1,27 @@
...
+
+  componentDidMount() {
+    const github = new GitHub()
+    github.getNotifications().
+           then(this.onNotificationsLoaded.bind(this)).
+           catch(this.onNotificationsError.bind(this))
+  }
+
+  onNotificationsLoaded(notifications) {
+    this.setState({ notifications })
```
cheshire137, line 19: I figure this will change when your Redux branch lands, @probablycorey.

test/models/fetcher-test.js
```diff
@@ -0,0 +1,17 @@
...
+describe('Fetcher', () => {
+  describe('getStatus', () => {
+    it('includes status and statusText properties', () => {
+      const fakeResponse = {
+        status: '404',
+        statusText: 'Not Found',
+      }
+      const expected = '404 Not Found'
+      const github = new GitHub()
+      assert(expected === github.getStatus(fakeResponse))
```
probablycorey: I think you can use `assert.deepEquals` instead of `assert` for this.

test/models/github-test.js
```diff
@@ -0,0 +1,15 @@
+const assert = require('assert')
+const fs = require('fs')
+const path = require('path')
+
+const GitHub = require('../../src/models/github')
+
+describe('GitHub', () => {
+  describe('getToken', () => {
+    it('returns contents of .env file', () => {
+      const tokenPath = path.join(__dirname, '..', '..', '.env')
```
probablycorey, line 10: What do you think about faking this, mocking it somehow? Forcing the test envs to have an `.env` file seems uncecessary.

probablycorey:
This looks good! For the actual app I think the notifications API won't be powerful enough, but I think we can get what we need from search API 😄

test/models/github-test.js
cheshire137, line 10: I'm not opposed to the idea of faking it somehow. The .env file being used here isn't one specific to tests, though, it's using the same .env file the app itself uses to run. One of the package.json scripts forces a .env to exist, even if it's blank, as part of initial setup.

PR #18, opened 2016-07-23
Fix for older node versions, fix for Windows

### LINKED ISSUE #17: Get it to work with node <6
@chrismwendt reports that node 5.9.0 results in an error:

```
> npm start
​
> gh-notifications-snoozer@0.0.1 start /Users/chrismwendt/gh-notifications-snoozer
> ./scripts/verify-token && electron src/main.js
​
/Users/chrismwendt/gh-notifications-snoozer/src/models/github.js:6
class GitHub extends Fetcher {
^^^^^
​
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (/Users/chrismwendt/gh-notifications-snoozer/scripts/verify-token:3:16)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
```

It'd be cool if we didn't require node 6+.

cheshire137:
This branch fixes two things:
- Fixes #17 by adding `'use strict'` to models.
- Makes the app start in Windows.

/cc @probablycorey


probablycorey:
Ahh, this fixes the problem I was having with the verify script https://github.com/cheshire137/gh-notifications-snoozer/pull/14/commits/3820508873ca3b4b6ad24262d4f953c29b9674ce

But I don't think the verify-script is needed anymore though, we can show that error directly in the app.

Issue #2, opened 2016-07-12
Allow adding rules


cheshire137:
I want to add a rule like "mentions `org/team-name`", "mentions `user`", or "created by `user`". These rules would be saved per user.

PR #20, opened 2016-07-23
Persistent storage for rules

cheshire137:
This is a start for #2 to allow the user to save rules for filtering their notifications. This adds a Rule model that allows saving and loading rules by name. This branch is based on #18, so that should land first. This also adds a Rules model for managing the names of saved rules, so we can show a big list for the user to edit/delete/select one.

/cc @probablycorey


zeke:
It looks like you've got this all implemented all ready, but figured I'd let you know about [`electron-config`](https://github.com/sindresorhus/electron-config), a nice alternative to `electron-json-storage`. It has a synchronous API which keeps your code a bit cleaner and easier to reason about.

PR #21, opened 2016-07-23
Add test for GitHub getNotifications

cheshire137:
I meant to add this earlier before #16 landed.

/cc @probablycorey


probablycorey:
LGTM!

PR #23, opened 2016-07-30
GitHub api plus redux

probablycorey:
This PR does a few things:
1. It queries the search api for all issues and PRs that mention team Panda Rocket (we can make the types of queries customizable in another PR)
2. It changes the redux state to be an object that contains tasks, instead of an array of tasks. I did this so we will be able to store more than just tasks in the state. In retrospect, I should have waited until we actually needed this functionality before I implemented it.

What's next?
- Get updates via the API. The app should be able to query the GitHub API for new and updated issues since our last fetch. This is going to require us to store when we last did the query and also use that date in the `getPandaRocketTasks` method.
- We need to persist this data, probably using https://github.com/cheshire137/gh-notifications-snoozer/pull/20. I think using Redux middleware might be the best way to do this?

cc @cheshire137


src/models/github.js
```diff
@@ -11,6 +11,22 @@ class GitHub extends Fetcher {
     return this.get('notifications')
   }
 
+  getPandaRocketTasks() {
+    const urlPath = 'search/issues?q=team=identity-team-panda-rocket'
```
cheshire137: What about storing `identity-team-panda-rocket` in the JSON config file for now, as a more temporary way of specifying the team? This method could be called `getTeamTasks` and take a `teamName` that it passes in the query.

`const Config = require('../config.json')`

src/models/github.js
```diff
@@ -11,6 +11,22 @@ class GitHub extends Fetcher {
...
+    return this.get(urlPath).then(({ items }) => {
+      return items.map(item => {
+        return {
+          title: item.title,
+          body: item.body,
+          state: item.state,
+          createdAt: new Date(item.created_at),
```
cheshire137, line 24: Nice! This must be that mapping you talked about wanting to do to convert snake_case into camelCase.

test/reducers/reducer-test.js
```diff
@@ -0,0 +1,85 @@
+const assert = require('assert')
+const Redux = require('redux')
+
+const reducer = require('../../src/reducers/reducer')
+
+describe('reducers', () => {
+  it('has the correctd default initial state', () => {
```
cheshire137: Are we introducing something for systemd? :trollface: 

`correctd`

cheshire137:
Seems good though I'd prefer not starting with a hard-coded team name. :+1:

src/models/github.js
probablycorey, line 24: Yeah, I was going to make it automatic, but we only need a few fields so I did it manually.

src/models/github.js
probablycorey: I like that. Once we have a way to store the redux state locally, we can throw they team name in redux's state.

test/reducers/reducer-test.js
probablycorey: 😬

probablycorey:
@cheshire137 I changed the `getTask` call to be more generic, so any search params can be used. The reason I used a hard-coded team was as a temporary step, the final goal would be to have custom filters available. But putting that data in the config file works well too.

src/models/github.js
```diff
@@ -11,6 +11,24 @@ class GitHub extends Fetcher {
     return this.get('notifications')
   }
 
+  // https://developer.github.com/v3/search/#search-issues
+  getTasks(query=Config.searchQuery) {
```
cheshire137, line 15: Blowing my mind with this modern JavaScript! I didn't know about default argument values.

src/models/github.js
```diff
@@ -11,6 +11,24 @@ class GitHub extends Fetcher {
...
+    const urlPath = `search/issues?q=${query}`
```
cheshire137, line 16: We'll probably want an `encodeURIComponent` here later.

Issue #7, opened 2016-07-12
Allow modifying and deleting filters


cheshire137:
I should be able to edit a rule as well as delete it.

cheshire137:
Deletion is a thing, just need to edit rules/filters now.

PR #25, opened 2016-08-06
Add rule management

### LINKED ISSUE #2: Allow adding rules

cheshire137:
Fixes #2 and partially addresses #7. You can now create and delete rules, as well as view matching issues for each rule. @probablycorey I could use your help to get rid of `setState` calls in favor of the Redux way of managing state.

View and delete your filters:

screen shot 2016-08-06 at 5 55 23 pm

---

Add a filter:

screen shot 2016-08-06 at 5 54 47 pm

---

View notifications matching a rule:

screen shot 2016-08-06 at 5 27 45 pm

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -4,8 +4,14 @@ const React = require('react')
 const GitHub = require('../../models/github')
 const Filter = require('../Filter')
 const TaskList = require('../TaskList')
+const NewRule = require('../NewRule')
 
 class App extends React.Component {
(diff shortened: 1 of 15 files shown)


probablycorey:
I just tried this out and it is **AWESOME**!!!

I'm going to merge this and open up a new PR that tries to move the state setting into redux.

PR #26, opened 2016-08-20
Persist redux state

probablycorey:
This is step one in a multi-step processes. My goal is to have Redux be the single source of truth of our application state, and to have that state be automatically persisted.

Here are the steps:
1. Automatically persist the redux state so it can be used when the app starts up. This PR solves that.
2. Split our single reducer into multiple reducers so it can store the rules.
3. Remove all the state methods from react (e.g. `setState`) and rely on Redux for that.

cc @cheshire137


src/models/github.js
```diff
@@ -21,9 +21,9 @@ class GitHub extends Fetcher {
           title: item.title,
           body: item.body,
           state: item.state,
-          createdAt: new Date(item.created_at),
-          updatedAt: new Date(item.updated_at),
-          closedAt: item.closed_at ? new Date(item.closed_at) : null,
+          createdAt: item.created_at,
+          updatedAt: item.updated_at,
+          closedAt: item.closed_at ? item.closed_at : null,
```
cheshire137, line 26: This line's a little funny now that we're not doing the date parsing here. Just `item.closed_at` would suffice.

cheshire137:
👍

src/models/github.js
probablycorey, line 26: 😆

PR #30, opened 2016-08-27
Allow archiving tasks

### LINKED ISSUE #5: Allow archiving an issue/PR
From the list of issues and pull requests that match a given rule, I should be able to archive one such that it no longer shows in the list. This doesn't have to do anything with the issue/PR on GitHub, it just needs to be suppressed in this app.

It would be nice if there is activity on the issue/PR on GitHub that it would become un-archived in the app.

cheshire137:
- Fixes #5 by hiding a task from the list if it is archived.
- Persists archived and snoozed tasks so they remain archived/snoozed when you exit and restart the app.
- Adds a refresh button so you can fetch the latest issues and pull requests matching your filter from GitHub.
- Uses a pull request icon for pull requests, instead of an issue icon.
- Displays the repository each issue/pull request is coming from.

Here's a sample of what the persistent JSON storage file looks like:

```
% cat /Users/cheshire137/Library/Application\ Support/Electron/config.json
{
    "Mentions me": "is:open mentions:cheshire137",
    "rules": [
        "Mentions me"
    ],
    "issue-92485268": "2016-08-27T18:23:57.079Z",
    "archived": [
        "issue-92485268"
    ]
}
```

/cc @probablycorey

### DIFF
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -45,6 +45,15 @@ function snoozeTasks(tasks) {
   })
 }
 
+function archiveTasks(tasks) {
+  return tasks.map(task => {
+    if (task.isSelected) {
+      return Object.assign({}, task, { archive: true })
+    }
+    return task
+  })
+}
+
 function tasksReducer(tasks = [], action) {
   switch (action.type) {
     case 'TASKS_EMPTY':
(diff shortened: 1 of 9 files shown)


src/reducers/reducer.js
```diff
@@ -1,51 +1,132 @@
...
+      archivedTasks.push(task)
+      return Object.assign({}, task, { archivedAt })
+    }
+    return task
+  })
+  writeChanges(archivedTasks, ARCHIVED_KEY)
+  return updatedTasks
+}
+
+function tasksReducer(tasks = [], action) {
```
cheshire137, line 111: Renamed the function to avoid a warning about the `tasks` parameter having the same name as the `tasks` function.

cheshire137:
Landing this using @probablycorey's rule of "you can land it anyway if I haven't gotten around to looking at it."

In we gooooooo!

PR #32, opened 2016-08-28
Add menus, About page

### LINKED ISSUE #19: Have reasonable menus
As seen in Windows 10:

gh-notifications-snoozer-menu

We won't need most of those menus. A Help menu with 'About' and even 'Help' suboptions would be good.

cheshire137:
Fixes #19 by adding an application-wide menu. There seems to be a bug right now where the app thinks its name is 'Electron' even though 'name' and 'productName' are both specified in package.json. Likewise, it says the wrong version on the About page, coming from `app.getVersion()`.

This also sets the title of the window based on which view you're on.

/cc @probablycorey

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -1,4 +1,5 @@
 {
   "githubApiUrl": "https://api.github.com",
-  "searchQuery": ""
+  "searchQuery": "",
+  "bugReportUrl": "https://github.com/cheshire137/gh-notifications-snoozer/issues/new"
 }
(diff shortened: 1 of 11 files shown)


cheshire137:
I'm going to land this one @probablycorey, but I'd appreciate your thoughts on how we could fix the problem where the app thinks it's "Electron" of a different version than what we specify in package.json. It's being silly.

PR #47, opened 2016-09-04
Allow setting access token in the app

### LINKED ISSUE #44: Add ability to set auth token within the app
The instructions require creating a .env file with a personal GitHub access token in it in order to use the app. We could make this easier for the user by allowing them to paste in their access token or choose it from a file on their computer.

cheshire137:
Fixes #44 by adding a form in the app where the user can set their GitHub access token.

screen shot 2016-09-04 at 5 19 16 pm
screen shot 2016-09-04 at 5 19 04 pm

/cc @probablycorey


cheshire137:
Opened #52 as a way to further enhance this.

PR #67, opened 2016-09-11
Default filters for the user's issues, mentions, comments, and assign…

### LINKED ISSUE #45: Add some reasonable default filters
Since we're requiring the user to provide a personal access token, we can [find out their user name](https://developer.github.com/v3/users/#get-the-authenticated-user) and generate a few default filters to get them started. Ideas:
- **Open recent mentions** mentions:_user_ is:open sort:updated-desc
- **My popular open tasks** author:_user_ is:open interactions:>5 sort:updated-desc
- **My comments** commenter:_user_ sort:updated-desc
- **My open assignments** assignee:_user_ sort:updated-desc is:open

summasmiff:
…ments
Uses the login information to add some basic custom filters.
Fixes #45 

screen shot 2016-09-10 at 10 04 22 pm


src/components/App/App.jsx
```diff
@@ -67,6 +66,8 @@ class App extends React.Component {
     const github = new GitHub()
     github.getCurrentUser().then(user => {
       this.setState({ user })
+      DefaultFilters.init(user.login)
+      this.setState({ filters: Filters.findAll() })
```
cheshire137: Since this function is getting longer, maybe add a new method in App called `onUserLoad` and move lines 68-70 into it? Then you can `getCurrentUser().then(u => this.onUserLoad(u))` here.

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,58 @@
...
+const ElectronConfig = require('electron-config')
+const configName = process.env.NODE_ENV === 'test' ? 'config-test' : 'config'
+const storage = new ElectronConfig({ name: configName })
+const Filter = require('./Filter')
+const LastFilter = require('./LastFilter')
+const DEFAULT_FILTER_KEY = 'default-filters'
+
+class DefaultFilters {
+  static init(login) {
+    if (!storage.has(DEFAULT_FILTER_KEY)) {
```
cheshire137: I'd flip this to reduce nesting.

``` javascript
if (storage.has(DEFAULT_FILTER_KEY)) {
  return
}
```

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,58 @@
...
+      this.myIssuesFilter(login)
+      this.myCommentsFilter(login)
+      this.myAssignmentsFilter(login)
+      this.myMentionsFilter(login)
+      this.myPopularFilter(login)
+      this.save()
+    }
+    return
```
cheshire137: You can omit this.

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,58 @@
...
+    LastFilter.save(defaultFilter.key)
+  }
+
+  static myPopularFilter(login) {
+    const defaultFilter = new Filter('My Popular Issues')
+    defaultFilter.store(`author:${login} is:open interactions:>5 sort:updated-desc`)
+    LastFilter.save(defaultFilter.key)
+  }
+
+  static save() {
```
cheshire137: "save" doesn't seem quite right. Maybe `finishedAddingDefaultFilters`?

test/components/AppTest.js
```diff
@@ -79,9 +79,9 @@ describe('App', () => {
 
       const fetchedCalls = fetchMock.calls().matched
       assert.equal(2, fetchedCalls.length, 'Two fetch calls should be made')
-      assert(fetchedCalls[0][0].match(/\/search\/issues/),
-             'Fetch call should be to the user API')
-      assert(fetchedCalls[1][0].match(/\/user/),
+      assert(fetchedCalls[1][0].match(/\/search\/issues/),
+             'Fetch call should be to the search issues API')
```
cheshire137, line 83: Haha thanks. Lying failure messages...

test/models/default-filters-test.js
```diff
@@ -0,0 +1,62 @@
+const assert = require('assert')
```
cheshire137: Could you rename the file to DefaultFiltersTest.js?

test/models/default-filters-test.js
```diff
@@ -0,0 +1,62 @@
...
+const ElectronConfig = require('electron-config')
+const storage = new ElectronConfig({ name: 'config-test' })
+const DefaultFilters = require('../../src/models/DefaultFilters')
+const login = 'RupertScrivener'
```
cheshire137: O shit, this one's up for grabs, boys!

screen shot 2016-09-11 at 4 34 45 pm

test/models/default-filters-test.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe('DefaultFilters', () => {
+  describe('init', () => {
+    xit('calls each filter function and then the save function', () => {
+      // what is the preferred way to spy on functions?
```
cheshire137: Maybe instead of spying on what the function does, assert its side effects? Assert there's no default-filters key in storage, call init, then check that `storage.has('default-filters')` is true. Could also assert that `filters` in storage now has 'My Issues', etc. in it.

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,58 @@
...
+
+  static myIssuesFilter(login) {
+    const defaultFilter = new Filter('My Issues')
+    defaultFilter.store(`author:${login} is:open sort:updated-desc`)
+    LastFilter.save(defaultFilter.key)
+  }
+
+  static myCommentsFilter(login) {
+    const defaultFilter = new Filter('My Comments')
+    defaultFilter.store(`commenter:${login} sort:updated-desc`)
```
cheshire137: Since these myXFilter methods are very similar in code, we could maybe DRY things up by using an object.

``` javascript
const DEFAULT_FILTERS = {
  'My Issues': 'author:login is:open sort:updated-desc',
  'My Comments': 'commenter:login sort:updated-desc',
  ...
}

static addDefaultFilters() {
  const filterNames = Object.keys(DEFAULT_FILTERS)
  filterNames.forEach(name => {
    const filter = new Filter(name)
    const value = DEFAULT_FILTERS[name].replace('login', login)
    filter.store(value)
  })
  LastFilter.save(filterNames[0])
  storage.set(DEFAULT_FILTER_KEY, 'default-filters')
}
```

Then `init` can just call this method, or do it inline even. Thoughts on this?

cheshire137:
Seems legit! Thanks for adding tests, too. 😀 Left a few comments on naming and reducing duplication, but this looks solid! 👍

src/models/DefaultFilters.js
summasmiff: I was trying to think of ways of DRY-ing it up as I saw all those similar methods, but wondered how to keep the default filter list editable enough. I like this, it makes it clear what each filter is, and gives a good place to add more without making this file too long.

src/components/App/App.jsx
summasmiff: Sure! It seems weird that this needs to be included two places - post-auth check (if the user is logged in) and post-auth (if they're not). Ideas?

src/models/DefaultFilters.js
summasmiff: There seems to be a scoping issue in this block. My outer function has access to the `login` param, and returns the correct value. However inside the forEach loop, `login` is undefined. I've attempted to set `const userLogin = login` to create a standard var with it, but no luck there either. 

```
filterNames.forEach(name => {
    const filter = new Filter(name)
    const value = DEFAULT_FILTERS[name].replace('login', login) <<- UNDEFINED!
    filter.store(value)
  })
```

cheshire137: Are you doing the `filterNames.forEach` loop inside the init method that has the `login` parameter?

Could try converting DefaultFilters into a class with a `constructor(login)`, set `this.login = login` in it, and make the method doing the `filterNames.forEach` loop a regular instance method (drop the `static`). Then you could use `this.login` in it, and the `this` even inside the `forEach` block, since you're doing the new-style `=> {` function syntax, should refer to the instance of the class and not the `forEach` function.

summasmiff:
@cheshire137 good to go?

src/components/App/App.jsx
```diff
@@ -26,16 +25,22 @@ class App extends React.Component {
...
       if (key) {
         this.loadFilter(key)
       } else {
         this.manageFilters()
       }
-      this.loadUser()
     }
   }
 
+  onUserLoad(user) {
```
cheshire137, line 38: Maybe rename to `login` so it's clear you're not getting a full user object, just a string?

**Edit:** keep this `user` and just pass `user.login` on line 39, if you update the caller of onUserLoad to give it a full user.

src/components/App/App.jsx
```diff
@@ -67,6 +72,7 @@ class App extends React.Component {
     const github = new GitHub()
     github.getCurrentUser().then(user => {
       this.setState({ user })
+      this.onUserLoad(user.login)
     }).catch(error => {
```
cheshire137: Would you instead do:

``` javascript
github.getCurrentUser().then(u => this.onUserLoad(u)).`
```

and move the `this.setState({ user })` into `onUserLoad`?

src/components/App/App.jsx
```diff
@@ -123,6 +129,7 @@ class App extends React.Component {
 
   finishedWithAuth(user) {
     this.setState({ user })
+    this.onUserLoad(user.login)
```
cheshire137: Lines 131-132 can be reduced to a single line if you make `onUserLoad` expect a full user object and have it do the setState call.

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,40 @@
...
+
+  addDefaults() {
+    if (storage.has(DEFAULT_FILTER_KEY)) {
+      return
+    }
+
+    const filterNames = Object.keys(DEFAULT_FILTERS)
+    filterNames.forEach(name => {
+      const filter = new Filter(name)
+      const value = DEFAULT_FILTERS[name].replace(/login/g, this.login)
```
cheshire137, line 32: That's some future thinking right there, with the `g`. 👍

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,40 @@
...
+
+const ElectronConfig = require('electron-config')
+const configName = process.env.NODE_ENV === 'test' ? 'config-test' : 'config'
+const storage = new ElectronConfig({ name: configName })
+const Filter = require('./Filter')
+const LastFilter = require('./LastFilter')
+const DEFAULT_FILTER_KEY = 'default-filters'
+
+const DEFAULT_FILTERS = {
+  'My Issues': 'author:login is:open sort:updated-desc',
```
cheshire137: I think you can include `type:issue` to ensure the results are only issues, to better match the name 'My Issues'.

cheshire137: Could similarly add a 'My Pull Requests' filter for the user to keep track of when their PRs get new activity. `type:pr` for that one, according to https://help.github.com/articles/searching-issues/.

test/models/DefaultFiltersTest.js
```diff
@@ -0,0 +1,53 @@
...
+      const filters = new DefaultFilters(login)
+      filters.addDefaults()
+    })
+
+    it('creates the default filters', () => {
+      assert.equal(true, storage.has('My Issues'))
+      assert.equal(true, storage.has('My Comments'))
+      assert.equal(true, storage.has('My Assignments'))
+      assert.equal(true, storage.has('My Mentions'))
+      assert.equal(true, storage.has('My Popular Issues'))
```
cheshire137: I think just `assert(` instead of `assert.equal(true,`.

src/models/DefaultFilters.js
summasmiff: Lol you never know, might need to replace it more than once!

src/components/App/App.jsx
```diff
@@ -26,16 +25,23 @@ class App extends React.Component {
...
       }
-      this.loadUser()
     }
   }
 
+  onUserLoad(user) {
+    this.setState({ user })
+    const filters = new DefaultFilters(user.login)
+    filters.addDefaults()
+    this.setState({ filters: Filters.findAll() })
```
cheshire137: Could wait and combine both the new state properties. Remove line 39 and do `this.setState({ filters: Filters.findAll(), user })` here.

src/components/App/App.jsx
```diff
@@ -65,12 +71,12 @@ class App extends React.Component {
...
   loadUser() {
     const github = new GitHub()
-    github.getCurrentUser().then(user => {
-      this.setState({ user })
-    }).catch(error => {
-      console.error('failed to load user', error)
-      GitHubAuth.deleteToken()
-    })
+    github.getCurrentUser()
+        .then(user => this.onUserLoad(user))
```
cheshire137: I'd say either put the `.` on the preceding line and align `then` with `get`, or keep the `.` here and align it with `.get`.

src/models/DefaultFilters.js
```diff
@@ -0,0 +1,41 @@
...
+const ElectronConfig = require('electron-config')
+const configName = process.env.NODE_ENV === 'test' ? 'config-test' : 'config'
+const storage = new ElectronConfig({ name: configName })
+const Filter = require('./Filter')
+const LastFilter = require('./LastFilter')
+const DEFAULT_FILTER_KEY = 'default-filters'
+
+const DEFAULT_FILTERS = {
+  'My Issues': 'author:login is:open sort:updated-desc type:issue',
+  'My Pull Requests': 'author:login is:open sort:updated-desc type:pr',
```
cheshire137, line 12: Swanky!

src/models/DefaultFilters.js
cheshire137: I could see a filter being like `-author:cheshire137 commenter:cheshire137` to show things I've commented on but didn't open myself, for example.

test/models/DefaultFiltersTest.js
```diff
@@ -0,0 +1,59 @@
...
+      assert(storage.has('My Issues'))
+      assert(storage.has('My Pull Requests'))
+      assert(storage.has('My Comments'))
+      assert(storage.has('My Assignments'))
+      assert(storage.has('My Mentions'))
+      assert(storage.has('My Popular Items'))
+    })
+
+    it('saves the DEFAULT_FILTER_KEY', () => {
+      assert.equal(true, storage.has(DEFAULT_FILTER_KEY))
```
cheshire137:
``` diff
- assert.equal(true, 
+ assert(
```

src/components/App/App.jsx
summasmiff: Ok, I'll keep the . on the second line and align it with `.get`. I hate trailing dots because then if you need to edit it you have to edit two lines rather than one. I think it's rather neat to have your additions/deletions be one line rather than two where possible. The spacing is awkward since I think convention is to indent twice for newlines for chained methods? I like the align better.

cheshire137 (approved):
Thanks!

PR #68, opened 2016-09-11
Allow viewing and restoring ignored/archived/snoozed tasks

### LINKED ISSUE #28: Allow viewing snoozed, ignored, and archived tasks
When I snooze/archive/ignore a task, it's hidden immediately. I should be able to view tasks I've snoozed so I can restore them.

cheshire137:
Fixes #28.

### DIFF
--- a/src/components/HiddenTaskList/HiddenTaskList.jsx
+++ b/src/components/HiddenTaskList/HiddenTaskList.jsx
@@ -84,4 +84,6 @@ HiddenTaskList.propTypes = {
 
 const mapStateToProps = state => ({ tasks: state.tasks })
 
-module.exports = connect(mapStateToProps)(hookUpStickyNav(HiddenTaskList, 'hidden-task-list-navigation'))
+const stickyNavd = hookUpStickyNav(HiddenTaskList,
+                                   'hidden-task-list-navigation')
+module.exports = connect(mapStateToProps)(stickyNavd)
--- a/src/components/HiddenTaskList/HiddenTaskList.jsx
+++ b/src/components/HiddenTaskList/HiddenTaskList.jsx
@@ -1,4 +1,7 @@
 const React = require('react')
+const { connect } = require('react-redux')
+
+const HiddenTaskListItem = require('../HiddenTaskListItem')
 const hookUpStickyNav = require('../hookUpStickyNav')
 
 class HiddenTaskList extends React.Component {
@@ -20,6 +23,11 @@ class HiddenTaskList extends React.Component {
           </div>
         </nav>
         <div className="hidden-task-list-container">
+          <ol className="task-list">
+            {this.props.tasks.map(task =>
+              <HiddenTaskListItem {...task} key={task.storageKey} />
+            )}
+          </ol>
         </div>
       </div>
     )
@@ -27,7 +35,11 @@ class HiddenTaskList extends React.Component {
 }
 
 HiddenTaskList.propTypes = {
+  tasks: React.PropTypes.array.isRequired,
+  dispatch: React.PropTypes.func.isRequired,
   cancel: React.PropTypes.func.isRequired,
 }
 
-module.exports = hookUpStickyNav(HiddenTaskList, 'hidden-task-list-navigation')
+const mapStateToProps = state => ({ tasks: state.tasks })
+
+module.exports = connect(mapStateToProps)(hookUpStickyNav(HiddenTaskList, 'hidden-task-list-navigation'))
--- a/src/components/HiddenTaskList/HiddenTaskList.jsx
+++ b/src/components/HiddenTaskList/HiddenTaskList.jsx
@@ -3,6 +3,7 @@ const { connect } = require('react-redux')
 
 const HiddenTaskListItem = require('../HiddenTaskListItem')
 const hookUpStickyNav = require('../hookUpStickyNav')
+const TaskVisibility = require('../../models/TaskVisibility')
 
 class HiddenTaskList extends React.Component {
   onRestoreClick(event) {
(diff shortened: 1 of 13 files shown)


src/components/HiddenTaskList/HiddenTaskList.jsx
```diff
@@ -0,0 +1,89 @@
...
+    return (
+      <p className="content">
+        There are no issues or pull requests that have been snoozed, archived,
+        or ignored for this filter.
+      </p>
+    )
+  }
+
+  render() {
+    const { activeFilter } = this.props
```
summasmiff, line 50: Does this just set a const with the name of the current activeFilter that you could ordinarily call as `this.props.activeFilter`?

src/components/HiddenTaskList/HiddenTaskList.jsx
```diff
@@ -0,0 +1,89 @@
...
+    const hiddenTasks = this.props.tasks.filter(task => TaskVisibility.isHiddenTask(task))
+    return (
+      <div>
+        <nav id="hidden-task-list-navigation" className="top-nav nav">
+          <div className="nav-left">
+            <h1 className="title">
+              <a href="#" onClick={event => this.cancel(event)}>Tasks</a>
+              <span> / </span>
+              Hidden
+              <span className="subtitle"> in &ldquo;{activeFilter}&rdquo;</span>
```
summasmiff, line 60: This is cool, but without a way of switching what your activeFilter is (other than going back, changing the filter, and returning), I feel like this could be annoying. Since the hidden tasks are ...hidden, I might forget what filter would reveal them, and have to go on a wild goose chase to find them so I could un-hide them. We may want to add an 'All' to the default filters list.

Any thoughts on making the hidden task list show all hidden tasks, rather than filtered?

src/components/HiddenTaskListItem/HiddenTaskListItem.jsx
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+    return iconClasses.join(' ')
+  }
+
+  render() {
+    const { updatedAt, repository, title, repositoryOwner, user, storageKey,
+            url, state, repositoryOwnerAvatar, userAvatar } = this.props
+
+    return (
```
summasmiff, line 43: I wonder if this could be made more modular. I feel like we're using nearly the same render as the regular task-list-item component.

src/reducers/reducer.js
```diff
@@ -16,6 +16,7 @@ function getSavedTaskKeys(key) {
 function writeChanges(tasks, typeKey) {
   const existingTaskKeys = getSavedTaskKeys(typeKey)
   const newTaskKeys = tasks.map(task => task.storageKey)
+  console.info(typeKey, newTaskKeys)
```
summasmiff, line 19: 💀 ?

src/reducers/reducer.js
```diff
@@ -59,21 +73,29 @@ function updateTasks(tasks, action) {
...
-  return tasks.map(task => {
+  const selectedTasks = []
+  const updatedTasks = tasks.map(task => {
     if (task.storageKey === action.task.storageKey) {
+      selectedTasks.push(task)
       return Object.assign({}, task, { isSelected: true })
     }
     return task
   })
+  console.info('select', selectedTasks.map(task => task.storageKey))
```
summasmiff, line 84: 💀 ?

src/reducers/reducer.js
```diff
@@ -59,21 +73,29 @@ function updateTasks(tasks, action) {
...
-  return tasks.map(task => {
+  const deselectedTasks = []
+  const updatedTasks = tasks.map(task => {
     if (task.storageKey === action.task.storageKey) {
+      deselectedTasks.push(task)
       return Object.assign({}, task, { isSelected: false })
     }
     return task
   })
+  console.info('deselect', deselectedTasks.map(task => task.storageKey))
```
summasmiff, line 97: 💀 ?

src/reducers/reducer.js
```diff
@@ -116,14 +151,40 @@ function archiveTasks(tasks) {
...
+      restoredTasks.push(task)
+      return Object.assign({}, task, {
+        archivedAt: null,
+        ignore: false,
+        snoozedAt: null,
+      })
+    }
+    return task
+  })
+  console.info('restore', restoredTasks.map(task => task.storageKey))
```
summasmiff, line 181: 💀 ?

summasmiff (approved):
Looks good! Other than wanting to know your thoughts about why using filters for this, I think it's fine.

src/components/HiddenTaskList/HiddenTaskList.jsx
cheshire137: Yes! This line is equivalent to:

``` javascript
const activeFilter = this.props.activeFilter
```

This is part of Glorious Modern JavaScript ™️ so you can use it to suck out properties from objects wherever you like. This is also what's happening in GitHub.js on this line:

``` javascript
return this.get(urlPath).then(({ items }) => items.map(d => getTask(d)))
```

src/components/HiddenTaskList/HiddenTaskList.jsx
cheshire137: > Any thoughts on making the hidden task list show all hidden tasks, rather than filtered?

I agree that would be preferable. I didn't do that for a technical limitation right now: we don't know the details of these hidden tasks until we go fetching them from the GitHub API. So we have to switch to a filter to make the API call to get all the issues and PRs, and then we split those between the ones that are visible and the ones that are hidden. The only data we have about hidden tasks is their type (pull or issue) and their ID. Try running `cat /Users/your_user_name/Library/Application\ Support/Electron/config.json` locally to see what's getting stored in our JSON storage file.

src/reducers/reducer.js
cheshire137: I intentionally added some `console.info` calls thinking they could stick around. If you see a stray `console.log`, that was meant as temporary debugging and should be removed!

This is a pattern I've followed in other projects of mine, notably Chrome extensions. If you open the browser inspector, it's nice to see a log of the actions the app has been taking. Helps to reconstruct what may have gone wrong if there's a bug.

PR #69, opened 2016-09-14
Set newly created filters as LastFilter, update results accordingly

### LINKED ISSUE #61: After creating a filter, apply it immediately
When I create a filter and return to the task list, I should see my new filter as the selected one. Especially for the user's first filter, it should be selected in the filter menu and queried immediately to show matching tasks.

/cc @summasmiff

summasmiff:
Fixes #61 

When a new filter is created, sets it as the active filter and updates the Task List.

### DIFF
--- a/test/components/NewFilterTest.js
+++ b/test/components/NewFilterTest.js
@@ -1,15 +1,40 @@
 const assert = require('assert')
 const React = require('react')
 const ReactDOM = require('react-dom')
+const ReactRedux = require('react-redux')
+const Redux = require('redux')
+const fetchMock = require('fetch-mock')
 const TestUtils = require('react-addons-test-utils')
-const NewFilter = require('../../src/components/NewFilter/NewFilter.jsx')
+const NewFilter = require('../../src/components/NewFilter')
+const ElectronConfig = require('electron-config')
+const storage = new ElectronConfig({ name: 'config-test' })
+const reducer = require('../../src/reducers/reducer')
+
+function renderPage(store) {
+  // NewFilter calls loadFilter and save functions from the main App onSubmit.
+  return TestUtils.renderIntoDocument(
+    <ReactRedux.Provider store={store}>
+      <NewFilter
+        loadFilter={key => key}
+        save={() => 'saved'}
+      />
+    </ReactRedux.Provider>
+  )
+}
 
 describe('NewFilter', () => {
   let renderedDOM
+  let store
 
   before(() => {
-    const component = TestUtils.renderIntoDocument(<NewFilter />)
+    store = Redux.createStore(reducer)
+    fetchMock.mock('*', {})
+    const component = renderPage(store)
     renderedDOM = () => ReactDOM.findDOMNode(component)
+  })
+
+  after(() => {
+    fetchMock.restore()
   })
 
   it('shows the form', () => {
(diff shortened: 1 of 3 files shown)


test/components/NewFilterTest.js
```diff
@@ -0,0 +1,35 @@
+const assert = require('assert')
+const React = require('react')
+const ReactDOM = require('react-dom')
+const TestUtils = require('react-addons-test-utils')
+const NewFilter = require('../../src/components/NewFilter/NewFilter.jsx')
```
cheshire137: You should be able to just `require('../../src/components/NewFilter')` if that directory has a package.json file in it like the other package.jsons (excluding the root one).

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,35 @@
...
+      const saveButton = renderedDOM().querySelectorAll('#new-filter-save')
+      assert.equal(1, saveButton.length)
+    })
+
+    it('has required form fields', () => {
+      const valueInput = renderedDOM().querySelectorAll("input[name='filterValue']")
+      assert.equal(1, valueInput.length)
+    })
+
+    xit('sets a created filter as LastFilter', () => {
```
cheshire137: To be more concise, you can also just do `it('sets a created filter as LastFilter')` and it will count as a pending test.

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,35 @@
...
+
```
summasmiff, line 60: I would like to do something like  

```
filterKey.inputText('butts')
filterValue.inputText('sortby:butts')
saveButton.click()
```

or likewise here to test the creation of a filter and then look at the storage key in this test, but I don't know how to

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  it('shows the form', () => {
+    const newFilterForm = renderedDOM().querySelectorAll('.new-filter-form')
+    assert.equal(1, newFilterForm.length)
+  })
+
+  describe('Create', () => {
+    it('has a save button', () => {
+      const saveButton = renderedDOM().querySelectorAll('#new-filter-save')
```
cheshire137: Should be able to do `querySelector` and then `assert(saveButton)`. `querySelectorAll` returns an array while `querySelector` returns the single node that matches.

cheshire137 (approved):
Thanks for adding tests for NewFilter!

test/components/NewFilterTest.js
cheshire137: You might find https://facebook.github.io/react/docs/test-utils.html helpful for the button click. See https://github.com/cheshire137/gh-notifications-snoozer/blob/d9808a818531218355634cb60fc0ccb95366024c/test/components/TaskListTest.js#L89 for a place we used that.

summasmiff:
@cheshire137 I have addressed the comments and added a test for the NewFilter submit button!

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,67 @@
...
+    renderedDOM = () => ReactDOM.findDOMNode(component)
+  })
+
+  after(() => {
+    fetchMock.restore()
+  })
+
+  it('shows the form', () => {
+    const newFilterForm = renderedDOM().querySelectorAll('.new-filter-form')
+    assert.equal(1, newFilterForm.length)
```
cheshire137: You can use `querySelector` and then just `assert(newFilterForm)` here if you like.

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,67 @@
...
+  })
+
+  describe('Create', () => {
+    it('has a save button', () => {
+      const saveButton = renderedDOM().querySelector('#new-filter-save')
```
cheshire137: Rather than having an ID on the button and asserting it exists by that, since an ID can be put on anything, maybe `querySelector('button[type="submit"]')`.

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,67 @@
...
+
+    it('has required form fields', () => {
+      const valueInput = renderedDOM().querySelector("input[name='filterValue']")
+      assert(valueInput)
+    })
+
+    it('sets a created filter as LastFilter', () => {
+      const filterForm = renderedDOM().querySelector('form.new-filter-form')
+      const valueInput = renderedDOM().querySelector("input[name='filterValue']")
+      const filter = 'author:LuluPopplewell'
```
cheshire137, line 59: You're a font of awesome and available GitHub names!

screen shot 2016-09-21 at 6 12 29 pm

test/components/NewFilterTest.js
```diff
@@ -0,0 +1,67 @@
...
+
+      valueInput.value = filter
+      TestUtils.Simulate.change(valueInput)
+      TestUtils.Simulate.submit(filterForm)
+      assert.equal(storage.get(filter), 'author:LuluPopplewell')
```
cheshire137: Maybe flip the ordering here, so the first parameter is the expected value and the second is the actual value. You could also add a code comment above this line about how if a filter is unnamed, it gets its value as its name.

You could also reuse the `filter` variable here instead of typing author:LuluPopplewell again.

Also, I think you want another assertion here that this new filter is the last filter: `assert.equal(filter, storage.get('last-filter'))`.

cheshire137 (approved):

test/components/NewFilterTest.js
summasmiff: Sounds good! I'm going to leave 'LuluPoppelwell' though, since A. I like that name and 2) I think it's more clear to see what's being tested than `assert.equal(filter, storage.get(filter))`

Issue #38, opened 2016-09-03
Add snooze/ignore/archive buttons to each task


cheshire137:
So I can snooze/ignore/archive a task with one click instead of two (select, click a button), it would be nice to have a snooze/ignore/archive button on each task list item. Could use an icon/emoji to represent each action.

😴 = snooze

📥 = archive

❌ = ignore

cheshire137:
I think this would get a little unwieldy. I think adding #39 instead will be sufficient.

Issue #36, opened 2016-09-03
Fix imported LESS not rebuilding when it's changed


cheshire137:
This is dependent on https://github.com/electron/electron-compile/issues/27 so will probably require a version bump in package.json to fix.

PR #71, opened 2016-09-15
Add tabbed navigation

summasmiff:
Creates a primary tabbed navigation for the app.
Adds a secondary navigation for certain screens if they need additional functionality. 

~~Everything is working fine, but I want to refactor the main render function in App.jsx to reduce some of the duplication.~~ Done!

~~Needs some tests~~ Done!
screen shot 2016-09-22 at 3 16 19 pm

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -52,6 +52,51 @@ class App extends React.Component {
     })
   }
 
+  getViewContents() {
+    switch (this.state.view) {
+      case 'tasks': return (
+        <TaskList
+          addFilter={() => this.showNewFilterForm()}
+          changeFilter={key => this.loadFilter(key)}
+          manageFilters={() => this.manageFilters()}
+          user={this.state.user}
+          showAuth={() => this.showAuth()}
+        />)
+      case 'filters': return (
+        <FilterList
+          filters={this.state.filters}
+          delete={key => this.deleteFilter(key)}
+          edit={key => this.editFilter(key)}
+          addFilter={() => this.showNewFilterForm()}
+          cancel={() => this.showTaskList()}
+        />)
+      case 'edit-filter': return (
+        <EditFilter
+          filter={this.state.filter}
+          save={() => this.savedFilter()}
+          addFilter={() => this.showNewFilterForm()}
+          cancel={() => this.manageFilters()}
+          delete={key => this.deleteFilter(key)}
+        />)
+      case 'about': return (
+        <About cancel={() => this.showTaskList()} />)
+      case 'new-filter': return (
+        <NewFilter
+          save={() => this.savedFilter()}
+          cancel={() => this.showTaskList()}
+          manageFilters={() => this.manageFilters()}
+        />)
+      // Auth is default to ensure token
+      default: return (
+        <Auth
+          done={user => this.finishedWithAuth(user)}
+          isAuthenticated={GitHubAuth.isAuthenticated()}
+          user={this.state.user}
+        />
+      )
+    }
+  }
+
   showAbout() {
     ipcRenderer.send('title', 'About')
     this.changeView('about')
@@ -132,51 +177,6 @@ class App extends React.Component {
   finishedWithAuth(user) {
     this.onUserLoad(user)
     this.showTaskList()
-  }
-
-  getViewContents() {
-    switch (this.state.view) {
-      case 'tasks': return (
-        <TaskList
-          addFilter={() => this.showNewFilterForm()}
-          changeFilter={key => this.loadFilter(key)}
-          manageFilters={() => this.manageFilters()}
-          user={this.state.user}
-          showAuth={() => this.showAuth()}
-        />)
-      case 'filters': return (
-        <FilterList
-          filters={this.state.filters}
-          delete={key => this.deleteFilter(key)}
-          edit={key => this.editFilter(key)}
-          addFilter={() => this.showNewFilterForm()}
-          cancel={() => this.showTaskList()}
-        />)
-      case 'edit-filter': return (
-        <EditFilter
-          filter={this.state.filter}
-          save={() => this.savedFilter()}
-          addFilter={() => this.showNewFilterForm()}
-          cancel={() => this.manageFilters()}
-          delete={key => this.deleteFilter(key)}
-        />)
-      case 'about': return (
-        <About cancel={() => this.showTaskList()} />)
-      case 'new-filter': return (
-        <NewFilter
-          save={() => this.savedFilter()}
-          cancel={() => this.showTaskList()}
-          manageFilters={() => this.manageFilters()}
-        />)
-      // Auth is default to ensure token
-      default: return (
-        <Auth
-          done={user => this.finishedWithAuth(user)}
-          isAuthenticated={GitHubAuth.isAuthenticated()}
-          user={this.state.user}
-        />
-      )
-    }
   }
 
   render() {
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -9,6 +9,7 @@ const AppMenu = require('../../models/AppMenu')
 const GitHubAuth = require('../../models/GitHubAuth')
 const LastFilter = require('../../models/LastFilter')
 
+const TabbedNav = require('../TabbedNav')
 const TaskList = require('../TaskList')
 const FilterList = require('../FilterList')
 const NewFilter = require('../NewFilter')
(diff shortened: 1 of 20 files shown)


cheshire137:
So excite! 😄 How do you foresee this incorporating #68? Viewing your hidden tasks doesn't seem a primary thing users will want to do, it's more for when you accidentally click the wrong thing and want to restore something. So it needs to be accessible but not prominently.

Figuring out a good way to get at your hidden tasks when both branches land can come in a followup branch, it doesn't have to happen in here before this lands.

src/components/App/App.jsx
```diff
@@ -134,66 +180,17 @@ class App extends React.Component {
...
-    }
-
-    if (this.state.view === 'tasks') {
-      return (
-        <TaskList
-          addFilter={() => this.showNewFilterForm()}
-          changeFilter={key => this.loadFilter(key)}
+    return (
+      <div>
+        <TabbedNav
```
cheshire137, line 202: Sweet Baby Ray's... ❤️

src/components/App/App.less
```diff
@@ -8,22 +8,31 @@
 @import "../TaskList/TaskList.less";
 @import "../TaskListItem/TaskListItem.less";
 @import "../Auth/Auth.less";
+@import "../TabbedNav/TabbedNav.less";
 
 body {
-  background-color: #fff;
+  background-color: white;
```
cheshire137: Lol I can tell you've been working around #36.

src/components/App/App.less
```diff
@@ -8,22 +8,31 @@
...
   border-bottom: 1px solid rgba(211, 214, 219, 0.5);
 
-  .button.is-link {
+  .button.is-link, .label {
     text-decoration: none;
   }
 }
+
+nav#tabbed-nav {
+  margin-bottom: 0 !important;
```
cheshire137: Kind of a CSS code smell to see `!important`. Can you more specifically target the element to avoid this? Or maybe you're having to use this to work around an `!important` in the Bulma framework.

src/components/EditFilter/EditFilter.less
```diff
@@ -7,5 +7,5 @@
 }
 
 .edit-filter-container {
-  padding: 0 20px 20px 20px;
+  padding: 10px 20px 20px 20px;
```
cheshire137: Doesn't have to happen in this branch, but we use this 10/20/20/20 in enough views we could pull it into a common class to apply to each of those elements.

src/components/FilterListItem/FilterListItem.jsx
```diff
@@ -21,15 +21,15 @@ class FilterListItem extends React.Component {
               className="button is-link"
               title="Edit filter"
             >
-              <span className="octicon octicon-pencil"></span>
+              ✏️
```
cheshire137, line 24: Oh duh. It hadn't occurred to me that we use native emoji for icons in some places and Octicons in others.

I think the Octicons still make sense for GitHub-specific stuff, but I like using the native emoji for these general icons.

src/components/FilterListItem/FilterListItem.jsx
```diff
@@ -21,15 +21,15 @@ class FilterListItem extends React.Component {
...
             </button>
             <button
               onClick={() => this.props.delete(this.props.filter)}
               type="button"
               className="button is-link"
               title="Delete filter"
             >
-              <span className="octicon octicon-trashcan"></span>
+              ❌
```
cheshire137, line 32: Watch it: I think this is the emoji used for 'ignore' on the main screen. I don't necessarily like it for ignore, but I think they should be different for delete v. ignore.

src/components/TabbedNav/TabbedNav.jsx
```diff
@@ -0,0 +1,42 @@
+const React = require('react')
+const hookUpStickyNav = require('../hookUpStickyNav')
+
+class TabbedNav extends React.Component {
+  render() {
+    return (
+      <nav id="tabbed-nav" className="top-nav nav tabs is-toggle is-fullwidth">
+        <ul>
+          <li className={this.props.active === 'tasks' ? 'is-active' : ''}>
+            <a onClick={this.props.showTasks}>
```
cheshire137: Since there's no `href` necessary, I'd use `<button type="button"` for these.

src/components/TabbedNav/TabbedNav.less
```diff
@@ -0,0 +1,5 @@
+.tab {
+  padding-right: 5px;
+}
+
+
```
cheshire137: 💀

src/components/App/App.less
summasmiff: Yes, this was to work around Bulma. I may be able to remove this, I think I found a specific enough selector to not need it anymore.

src/components/FilterListItem/FilterListItem.jsx
summasmiff: Since this is GitHub specific app, it might be best to use Octicons for everything

src/components/FilterListItem/FilterListItem.jsx
summasmiff: It is, but I prefer an 'X' for delete rather than ignore...which raises a bigger question to me...what is the difference between snooze and ignore?

cheshire137: Snooze = hide this task for 24 hours. Ignore = never show me this task again.

summasmiff:
Re: #68, I think we can replace the "With selected: Snooze \* Archive \* Delete" with the "View: Snoozed \* Archived \* Deleted" after we move the snooze / archive / delete to be inline with notifications ( #38 ).

Ideally, the secondary nav should be for operations that aren't specific to any singular item but are sort of general to the current page, eg, add a new filter to the FilterList. Individual operations should all be inline with the item they affect.

src/components/FilterListItem/FilterListItem.jsx
summasmiff: Re: wanting ignore vs. delete to be different icons - I think in the context of this app, they are basically the same functionality - "I don't want to see this in my list". The notifications are from GitHub, so we don't want to delete them outright, but the only real difference is that a filter can't be un-deleted. 

so... unless there's a stronger icon than X, we should reserve X for things that are permanent. Archiving is a less permanent decision, so its icon should be less severe. Ideas for that icon?

src/main.js
```diff
@@ -22,7 +22,7 @@ function createWindow() {
   })
   mainWindow.loadURL(`file://${__dirname}/index.html`)
 
-  // mainWindow.webContents.openDevTools()
+  mainWindow.webContents.openDevTools()
```
summasmiff: whoops, this wasn't supposed to be here

src/components/TabbedNav/TabbedNav.jsx
summasmiff: I disagree - switching it to a button adds a lot of button styling I would have to override, only to make it more semantically correct. Using a `href`-less `<a>` is not ideal, but it works without outside modification.

cheshire137: Look again, your button is now `is-link`! 💎  See http://bulma.io/documentation/elements/button/ for the `is-link` class Bulma provides for buttons. I think using `button is-link` together on a button here would make it look basically right.

summasmiff:
Hey @cheshire137 : any thoughts about the use of the 'X' button? I think I've addressed the rest of your comments!

src/components/FilterListItem/FilterListItem.jsx
cheshire137: Maybe 🗑 for delete?

summasmiff: Doesn't look like it has universal support...
screen shot 2016-09-21 at 4 04 27 pm

cheshire137: Haha okay clearly not that emoji if it's not part of the standard set. I'm in El Capitan and that shows up as a trash can for me:

screen shot 2016-09-21 at 6 21 48 pm

Keep this X for now!

cheshire137:
Ooh now you got conflicts, son. Sorry about the merge!

summasmiff:
Do we want to call them "Notifications" or "Tasks"?

cheshire137:
User-facing I think they should always be tasks. In the code, some things are notifications because they come from the GitHub notifications API. Most of the data is coming from the issues API and they aren't actual notifications, they're just issues and PRs that the user is searching for via their filters. But many of these issues/PRs are tied to open notifications on GitHub for the user, so we mark those notifications as read, if they exist, once the user archives/ignores/snoozes a task.

summasmiff:
Cool. Updating this to use the "task" nomenclature in the UI exclusively.

src/components/EditFilter/EditFilter.less
summasmiff: Now you can use `.view-container` for this formatting!

src/components/App/App.jsx
```diff
@@ -55,6 +56,57 @@ class App extends React.Component {
     })
   }
 
+  getViewContents() {
+    switch (this.state.view) {
+      case 'tasks': return (
+        <TaskList
```
cheshire137, line 62: Seeing all these components listed closely like this will help with refactoring later. I think it'd be nice to reduce the duplication between method handlers passed to each component, like addFilter and manageFilters.

src/components/App/App.jsx
```diff
@@ -90,14 +142,14 @@ class App extends React.Component {
   }
 
   savedFilter() {
-    ipcRenderer.send('title', 'Notifications')
+    ipcRenderer.send('title', 'Tasks')
```
cheshire137, line 145: It'd be cool if we had changeView set the title based on the view to be displayed. That can come in the future, though.

src/components/TabbedNav/TabbedNav.jsx
```diff
@@ -0,0 +1,42 @@
+const React = require('react')
+const hookUpStickyNav = require('../hookUpStickyNav')
+
+class TabbedNav extends React.Component {
+  render() {
+    return (
+      <nav id="tabbed-nav" className="top-nav nav tabs is-toggle is-fullwidth">
+        <ul>
+          <li className={this.props.active === 'tasks' ? 'is-active' : ''}>
```
cheshire137, line 9: Could do `const { active } = this.props` just above the `return` on line 6, to shorten these lines.

src/components/TabbedNav/TabbedNav.jsx
```diff
@@ -0,0 +1,42 @@
...
+              <span>Tasks</span>
+            </a>
+          </li>
+          <li className={this.props.active === 'filters' ? 'is-active' : ''}>
+            <a id="filters-link" onClick={this.props.manageFilters}>
+              <span className="tab octicon octicon-beaker"></span><span>Filters</span>
+            </a>
+          </li>
+          <li className={this.props.active === 'auth' ? 'is-active' : ''}>
+            {typeof this.props.user === 'object' ? (
```
cheshire137, line 21: Could this `typeof` check be moved outside the `<li` so that the list item is only included if we do have a user? That might screw up the appearance. Just thinking we don't need an empty li sitting around.

test/components/AppTest.js
```diff
@@ -39,6 +39,12 @@ describe('App', () => {
       assert(ReactDOM.findDOMNode(appComponent))
     })
 
+    it('has a tabbed nav', () => {
+      const appComponent = renderPage(store)
+      const app = ReactDOM.findDOMNode(appComponent)
+      assert(app.querySelectorAll('tabbed-nav'))
```
cheshire137: This doesn't look right. I think you mean `assert(app.querySelector('.tabbed-nav'))` or even `assert.equal(1, app.querySelectorAll('.tabbed-nav').length)`.

test/components/TabbedNavTest.js
```diff
@@ -0,0 +1,24 @@
...
+    const component = TestUtils.renderIntoDocument(<TabbedNav />)
+    renderedDOM = () => ReactDOM.findDOMNode(component)
+  })
+
+  it('has three links', () => {
+    const notificationsLink = renderedDOM().querySelectorAll('notifications-link')
+    const filtersLink = renderedDOM().querySelectorAll('filters-link')
+    const authLink = renderedDOM().querySelectorAll('auth-link')
+
+    assert(notificationsLink)
```
cheshire137: Same problem as my comment above. `querySelectorAll` returns an array, and an empty array is always truthy. You're going to get nothing but empty arrays back when your selector is `'auth-link'`, for example, because that's not a CSS selector. I think you want `.auth-link` to find the element with that class, or `#auth-link` to find the element with that ID.

screen shot 2016-09-22 at 2 37 53 pm

I'd use `querySelector` which will return the first matching element or null, and a null is not truthy so the `assert` will fail as desired.

test/components/AppTest.js
summasmiff: I meant `querySelector`. My bad!

cheshire137:
I updated your tests! 🙆

cheshire137:
In we gooooo! Thanks! 😀

PR #76, opened 2016-09-24
Disable buttons when none selected, remove horizontal scrollbars

cheshire137:
- Disable 'Tasks' and 'Filters' top nav buttons when the user has not provided an access token.
- Make top navigation area more compact vertically.
- Get rid of horizontal scrollbar around filter select menu.
- Disable 'Snooze', 'Archive', and 'Ignore' buttons when no visible tasks are selected.
- Disable 'Restore' button when no hidden tasks are selected.
- Use an emoji for the refresh button icon.
- Spell out 'View hidden' instead of using an eye Octicon.
- Move 'View hidden' button below filter select menu.
- Get rid of rounded corners in top navigation, as well as borders adjacent to the window edges on the top nav buttons.
- Display page titles for edit filter, new filter, and 'manage filters' views.
- Make 'Authenticate' view less noisy by removing green styling on 'you are signed in' message.

screen shot 2016-09-24 at 4 15 16 pm

screen shot 2016-09-24 at 4 15 28 pm

screen shot 2016-09-24 at 4 15 40 pm

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -90,6 +90,7 @@ class App extends React.Component {
           save={() => this.savedFilter()}
           cancel={() => this.showTaskList()}
           manageFilters={() => this.manageFilters()}
+          loadFilter={key => this.loadFilter(key)}
         />)
       case 'hidden': return (
         <HiddenTaskList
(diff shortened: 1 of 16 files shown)


summasmiff:
~~Looks like we lost the "Hidden Tasks" eye icon. I had moved it to be next to the Refresh icon on the secondary toolbar.~~

I see now that it is a text link. I can't read good today.

PR #78, opened 2016-09-25
Update to Bulma 0.2.0

cheshire137:
Looks like a new default font. The primary color that was blue before is now a sea green. It was a little too light to make the top nav tabs readable, so I switched to the `is-info` button color seen on http://bulma.io/documentation/modifiers/syntax/.

screen shot 2016-09-25 at 11 39 00 am


summasmiff (approved):
Did you add `is-link` in another PR to the top-nav?

I think we should add another issue to do a general color-and-font overhaul to get away from Bulma's defaults at some point in the future, but for now, this is a good change to increase readability.   

However, Bulma's defaults are pretty cool and not boring, so I think they can stay for a while.

cheshire137:
> Did you add is-link in another PR to the top-nav?

Probably. I think `is-link` with `button` is required to make it not have prominent borders. On hover though, `.is-link.button` still looked like a button.

PR #79, opened 2016-09-25
Make suggestions as you type a search query

### LINKED ISSUE #56: Add suggestions for filters
Slack has a very nice search suggester/autocompleter:

screen shot 2016-09-10 at 3 05 23 pm

It would be helpful to have this same functionality when editing or creating a filter.

cheshire137:
Fixes #56. This is a start to having suggestions while you type a filter. We could expand this to provide a helpful date picker for filters like `created:`, as well as suggesting user names, organizations, teams, and repositories based on the authenticated user.

gh-snoozer-filter-suggestions


src/components/App/App.jsx
```diff
@@ -182,16 +192,15 @@ class App extends React.Component {
...
+    this.setState({ filter: key }, () => {
       ipcRenderer.send('title', `Edit Filter ${key}`)
       this.changeView('edit-filter')
     })
   }
 
   changeView(view) {
     window.scrollTo(0, 0)
-    this.setState({ view })
+    this.setState({ view, previousView: this.state.view })
```
summasmiff, line 203: Isn't `this.state.view` already reset at this point?

src/components/FilterHelp/FilterHelp.jsx
```diff
@@ -1,132 +0,0 @@
...
-            <code>interactions</code> &mdash; Filter by reaction + comment
-            count, e.g., <code>interactions:&lt;10</code>.
-          </li>
-        </ul>
-      </div>
-    )
-  }
-}
-
-module.exports = FilterHelp
```
summasmiff, line 132: Hmm, so no more filter glossary? I liked having at least a cheat sheet I could look at, even if just to give me ideas of all the stuff I could filter by.

src/components/FilterSuggester/FilterSuggester.jsx
```diff
@@ -0,0 +1,306 @@
...
+    tip: 'number',
+    suggestions: [
+      { name: '', prefix: 'interactions:', tip: 'number' },
+      { name: '>', prefix: 'interactions:', tip: 'number' },
+      { name: '<', prefix: 'interactions:', tip: 'number' },
+      { name: '>=', prefix: 'interactions:', tip: 'number' },
+      { name: '<=', prefix: 'interactions:', tip: 'number' },
+    ],
+  },
+]
```
summasmiff, line 136: Should all these autocomplete objects be moved to their own file? Seems unwieldy to have it here.

src/components/FilterSuggester/FilterSuggester.jsx
```diff
@@ -0,0 +1,306 @@
...
+    super(props)
+    const value = props.value || ''
+    this.state = {
+      value,
+      previousValue: value,
+      suggestions: [],
+    }
+    this.onKeyPress = this.onKeyPress.bind(this)
+    this.onKeyUp = this.onKeyUp.bind(this)
+  }
```
summasmiff, line 149: Can you explain a little about what this `constructor` does? Does it just rebind the keypresses and reset the value so the autocomplete can change every time the user types a letter?

summasmiff (approved):
I definitely think we should expand this to suggest the user soon! See #80 .

src/components/App/App.jsx
cheshire137: It's not! Not until after `this.setState` on this lines run will `this.state.view` be changed.

src/components/FilterHelp/FilterHelp.jsx
cheshire137: You think it would be helpful? I thought it was a little overwhelming having the suggester with the docs beneath since they list the same things, the docs just goes into more depth. Maybe I can restore this page in a followup branch, as a standalone help page.

src/components/FilterSuggester/FilterSuggester.jsx
cheshire137: Not a bad point. Maybe called filterSuggestions.json.

src/components/FilterSuggester/FilterSuggester.jsx
cheshire137: I'm using [this guy's technique](https://medium.com/@ericclemmons/react-event-preventdefault-78c28c950e46#.vtjap6b1c) of getting around the fact that events in your React handlers are not native events, thus calling `preventDefault` on them doesn't do anything. I needed to override pressing Enter and Esc so they wouldn't do the default action in the input field. Enter would normally submit the form, I want it to not submit the form if the user is just pressing Enter to apply a suggestion. Esc kept clearing out the field, and it still does but now I restore it immediately; I think the 'clear the field' logic is coming from the react-autocomplete package I'm using, but I couldn't figure out how to turn it off.

For using the native addEventListener and removeEventListener, the same function has to be passed to removeEventListener as was passed to addEventListener, otherwise, it doesn't get unbound. But we need the handlers to know what `this` is (namely the instance of FilterSuggester they're working on), so we have to do this `bind(this)` to tie them to that instance. That lets us just pass `this.onKeyPress` to addEventListener/removeEventListener, instead of creating an anonymous function like `e => this.onKeyPress(e)` as the parameter.

src/components/FilterHelp/FilterHelp.jsx
summasmiff: Perhaps a shorter version, with some very common ones. If you are not familiar _at all_ with the filter syntax, it's not terribly useful to have it autocomplete, because you are just staring at a blank page with no idea what to do. A short paragraph such as "Filters use standard Github formatting, for example `is:issue author:{user} sort:updated-desc`. Click here for a full list of modifiers!"

summasmiff: Honestly, I just hate having the possibility of a totally empty page. Check out #81 as an idea of how we can use the default filters as the helpful hint combined with autocomplete.

Issue #90, opened 2016-10-01
Add menu option to open dev tools


cheshire137:
We have `// mainWindow.webContents.openDevTools()` in src/main.js because it's convenient to run the app with the developer console open while debugging. It'd be nicer if we could remove that line altogether and just provide a menu option in src/models/AppMenu.js to open the console.

darrylpargeter:
Hey @cheshire137, 
is anyone currently working on this issue? if no one is would it be possible to do it?

cheshire137:
Have at it! :) It's all yours.

darrylpargeter:
Thank you =]

Though quick question are you wanting the menu option appearing within one of the already established menu options or within its own e.g Tools?

cheshire137:
Probably a new menu, File/Edit/Help don't seem to fit with this option. I think Tools or View sounds good.

PR #92, opened 2016-10-01
Dev Tool menu options added to the newly created Tools menu

darrylpargeter:
Pull Request for issue #90 

Added a Tools menu and a menu item to toggle the dev tools as well as a short cut to do the same


cheshire137:
I just tried it out. This is so much nicer! Thank you!

PR #96, opened 2016-10-03
Remove storage from Reducer

probablycorey:
This PR does two things:
1. Removes all the storage and state calls from the reducer

All state is automatically saved via `AsyncNodeStorage` in index.jsx. So we never have to think about it!
1. Refreshing the task list doesn't empty the filter state

We need to remember if we snoozed/archived/ignored the tasks. I don't think there is a need to empty the list first.


cheshire137:
My comment before you go updating tests is does this seem to work when you manually test the app? So you archive an issue, exit the app, reopen it, refresh the list, and that same issue still does not appear because it's still archived. You click 'View hidden' and confirm that issue is listed there. Likewise for ignoring and snoozing tasks.

Are archivedAt and snoozedAt times getting persisted as well? We'll need those to know when to make a task visible again.

If all that still seems the same, this is a great change since it's simplifying things so much! I knew it felt off that we were using this auto-rehydrater logic for persisting state but we had a manual JSON storage, too. I like unifying the two so all changes persisted to disk are done so through only one tool.

probablycorey:
- [x] So you archive an issue, exit the app, reopen it, refresh the list, and that same issue still does not appear because it's still archived.
- [x] Are archivedAt and snoozedAt times getting persisted as well?
- [x] You click 'View hidden' and confirm that issue is listed there. Likewise for ignoring and snoozing tasks.

The hidden stuff doesn't work because we null that info out. But I think I can show all tasks without needing to null that info out.

probablycorey:
Nevermind what I said about the 'View hidden' stuff. I didn't understand what `TASKS_RESTORE` did. Now that I do I can verify that it works!

cheshire137:
Do we know the path to the file that AsyncNodeStorage uses? I like periodically `cat`ing out the electron-config JSON file to check which tasks are snoozed, etc. for debugging.

test/reducers/reducerTest.js
```diff
@@ -203,26 +170,10 @@ describe('reducers', () => {
                    'selected task should not have archivedAt')
       assert.equal(dateStr, actual[1].archivedAt,
                    'deselected task should have archivedAt')
-
-      const archivedTaskKeys = storage.get('archived')
-      assert(archivedTaskKeys.indexOf('issue-99') < 0,
-             'task key should not be in the archived list')
```
cheshire137, line 209: This test still needs some check to ensure what was once archived is now restored. Maybe after the restore step, you assert that archivedAt is now null on the task?

test/reducers/reducerTest.js
```diff
@@ -243,17 +194,6 @@ describe('reducers', () => {
                    'selected task should not have snoozedAt')
       assert.equal(dateStr, actual[1].snoozedAt,
                    'deselected task should have snoozedAt')
-
-      const snoozedTaskKeys = storage.get('snoozed')
-      assert(snoozedTaskKeys.indexOf('issue-97') < 0,
```
cheshire137, line 248: Likewise, this is going away, but for the test name "restores snoozed, selected task" to be honest, we need to confirm that the task loses its snoozed status after restoring it. Maybe assert that after TASKS_RESTORE, snoozedAt is null again?

probablycorey:
> Do we know the path to the file that AsyncNodeStorage uses? I like periodically cating out the electron-config JSON file to check which tasks are snoozed, etc. for debugging.

You can find this out running `console.log(persistDir)` on https://github.com/cheshire137/gh-notifications-snoozer/blob/remove-state/src/index.jsx#L13

For me that becomes `/Users/probablycorey/Library/Application\ Support/Electron/reduxPersist%3Atasks`. It could use a better file name!

test/reducers/reducerTest.js
probablycorey: I think that is what line 169 does.

test/reducers/reducerTest.js
probablycorey: That's what line 193 does, right?

cheshire137:
Elaborating on what I said in Slack, if this doesn't work when you change filters, I think it's because we're persisting tasks as a whole here. So if issue-123 comes in via one filter, I snooze it, then I switch filters, issue-123 will be visible for that filter. That's because issue-123 has a snoozedAt for the first filter, but not for the second.

Could we have another reducer that persists storageKey + archivedAt/snoozedAt/ignored status?

test/reducers/reducerTest.js
cheshire137: Ah, lines 242-245 do what I was looking for! Sorry I missed it.

test/reducers/reducerTest.js
cheshire137: Lines 202-205 do what I was looking for, never mind!

Issue #84, opened 2016-09-27
Get rid of horizontal scrollbar in hidden tasks view


cheshire137:
Long filter names can cause the left column to have a horizontal scrollbar:

![screen shot 2016-09-27 at 4 47 29 pm](https://cloud.githubusercontent.com/assets/82317/18893127/26e931ca-84d2-11e6-85d6-fb85a4f015ec.png)

I think using a class like `is-three-quarters` from Bulma -- see http://bulma.io/documentation/grid/columns/ -- will fix the horizontal scrollbar.

mattxwang:
I might be looking at this issue wrong (let me know if I am), but if you just don't want to have a scrollbar, you can give the element a width and `overflow-x:none;`. More info [here](http://www.w3schools.com/cssref/css3_pr_overflow-x.asp). In addition, you can force the text to wrap if you have a fixed-width element (e.g. the filter name div) using `word-wrap: break-word;`. [Here](http://stackoverflow.com/questions/5120313/wrapping-the-text-inside-div-css) is a good StackOverflow question explaining how to use it.

Good luck!

cheshire137:
Setting `overflow-x: none` would remove the scrollbar, but we have a little more horizontal real estate that we're not using well right now. The top area is split 50/50 between the title area that's getting cut off and a single button on the right. The button on the right doesn't need as much room as it's got right now, while the title area could use it so we don't have to 1) cut the text off or 2) force wrapping. 👍

PR #97, opened 2016-10-04
Added is-8 class for hidden tasks filter container

s4kh:
It seems that is-8 class is more suitable than is-three-quarter. Fixes https://github.com/cheshire137/gh-notifications-snoozer/issues/84
selection_001

### DIFF
--- a/src/components/App/App.less
+++ b/src/components/App/App.less
@@ -78,3 +78,7 @@ body {
 .view-container {
   padding: 10px 20px 20px 20px;
 }
+
+.lh {
+  line-height: 32px;
+}
--- a/src/components/HiddenTaskList/HiddenTaskList.jsx
+++ b/src/components/HiddenTaskList/HiddenTaskList.jsx
@@ -37,8 +37,8 @@ class HiddenTaskList extends React.Component {
     return (
       <div>
         <nav id="hidden-task-list-navigation" className="secondary-nav nav">
-          <div className="is-8">
-            <h2 className="subtitle nav-item">
+          <div>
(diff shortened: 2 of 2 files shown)


cheshire137:
Hmm, I'm not sure the `is-8` in that context is working:

screen shot 2016-10-04 at 10 09 17 am

We may not be able to use the Bulma grid column classes with their nav bar classes, like `nav-left`. We could switch off of using the `nav-left`, `nav`, and `nav-right` classes specifically for the 'hidden tasks' view, in favor of [Bulma's `columns` classes](http://bulma.io/documentation/grid/columns/).

s4kh:
@cheshire137:
I don't think we need to remove all the `nav` classes. Removing the `nav-left` did the trick for me.

cheshire137:
Sorry for the delay! I just tried this out and I agree the nav-left isn't necessary. I think you can remove the `is-8` now, since it doesn't seem to be applying any styles:

screen shot 2016-10-10 at 3 15 18 pm

---

The h2 on the left is no longer vertically centered without the nav-left. I tried adding a `line-height: 32px` to it to make it match the height of the 'Restore' button on the right:

screen shot 2016-10-10 at 3 15 30 pm

---

Would you 1) remove `is-8` from the left div and 2) put a new class on that h2 and style it with `line-height: 32px`?

s4kh:
Done  :punch:

cheshire137:
Thanks! I tried it out, looks good. :)

cheshire137:
Also, 👊  to return your fist bump. 😀

PR #99, opened 2016-10-15
Paginate task list

### LINKED ISSUE #95: Hitting API rate limit
I'm hitting the GitHub API limit because of the notification look ups:

img

cheshire137:
Fixes #95 or at least makes it better by

1) Not fetching all pages of tasks immediately; and
2) Only fetching all pages of notifications once.

screen shot 2016-10-15 at 1 43 28 pm

### DIFF
--- a/src/models/GitHub.js
+++ b/src/models/GitHub.js
@@ -68,7 +68,8 @@ class GitHub extends Fetcher {
   getTasksFromUrl(urlPath) {
     return this.get(urlPath).then(result => {
       const { json, nextUrl } = result
-      return { tasks: json.items.map(d => getTask(d)), nextUrl }
+      return { tasks: json.items.map(d => getTask(d)), nextUrl,
+               currentUrl: urlPath }
     })
   }
 
--- a/src/models/GitHub.js
+++ b/src/models/GitHub.js
@@ -55,18 +55,22 @@ class GitHub extends Fetcher {
       date.setDate(date.getDate() - 31)
     }
     const dateStr = date.toISOString()
-    return this.get(`notifications?since=${encodeURIComponent(dateStr)}`)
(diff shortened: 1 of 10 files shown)


src/models/GitHub.js
```diff
@@ -55,18 +56,28 @@ class GitHub extends Fetcher {
...
-    const urlPath = `search/issues?q=${encodeURIComponent(query)}`
-    return this.get(urlPath).then(({ items }) => items.map(d => getTask(d)))
+    const params = `?per_page=${this.perPage}&q=${encodeURIComponent(query)}`
+    const url = `${Config.githubApiUrl}/search/issues${params}`
+    return this.getTasksFromUrl(url)
+  }
+
+  getTasksFromUrl(url) {
+    return this.get(url).then(result => {
+      const { json, nextUrl } = result
```
summasmiff, line 72: Can you explain how nextUrl works? Is this some baked in magic from the Github API?

I see how you're using the headers, and construction a url from a relative & absolute path, but I'm still lost on how it "knows" where to go next with the pagination.

summasmiff (approved):
Looks fine. I'm a little weirded out by having so much pagination logic in the main & task list, but I don't think there's any real need to isolate it as a generic pagination component, so ¯_(ツ)_/¯

src/models/GitHub.js
cheshire137: Yes! See https://developer.github.com/v3/#pagination. Each response that has a next page to it will provide a link to that next page of results in the header.

cheshire137:
> I'm a little weirded out by having so much pagination logic in the main & task list

If pagination comes up with the hidden task list or another component, I can pull it out into a separate module.

summasmiff:
Yeah, that's fine. I've just never been in a situation where we've only needed pagination on a single page until now.

PR #102, opened 2016-10-15
Use keytar for GitHub access token storage

### LINKED ISSUE #52: Use keytar to store GitHub access token
We can use the https://www.npmjs.com/package/keytar package instead of using our regular JSON storage for storing the access token. This way, if someone knows you're using this app, they won't know exactly which file to check to find your access token.

Hat tip 🎩  @probablycorey.

cheshire137:
Fixes #52 by storing the GitHub access token in secure storage based on the operating system, instead of in a plain JSON file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@ of interest to you.
 
 You must provide a GitHub personal access token once you start the app the first
 time. [Create a token](https://github.com/settings/tokens/new) with the `repo`
-scope.
+scope. When running the app, your token will be stored in Keychain in macOS,
+Gnome Keyring in Linux, and Credential Vault in Windows.
 
 ## How to Develop
 
--- a/README.md
+++ b/README.md
@@ -14,9 +14,18 @@ scope.
 
 ## How to Develop
 
-This was developed with npm version 3.10.3 and node version 6.3.1
+This was developed with npm version 3.10.3 and node version 6.3.1.
 
 ### First-time Setup
+
+In Linux, you may need to run
+
+```bash
+sudo apt-get install libgnome-keyring-dev
+```
+
+first because this app uses gnome-keyring. Afterward, including for macOS and
+Windows:
 
 ```bash
 # Clone this repository
--- a/README.md
+++ b/README.md
@@ -45,6 +45,18 @@ Run the app with:
 npm start
 ```
 
+### Troubleshooting
+
+> When I run `npm start`, I get "Error: Module version mismatch. Expected 49, got 48" in the JavaScript console.
+
+Exit the app and try running:
+
+```bash
+./node_modules/.bin/electron-rebuild
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -10,14 +10,24 @@ of interest to you.
...
 You must provide a GitHub personal access token once you start the app the first
 time. [Create a token](https://github.com/settings/tokens/new) with the `repo`
-scope.
+scope. When running the app, your token will be stored in Keychain in macOS,
+Gnome Keyring in Linux, and Credential Vault in Windows.
 
 ## How to Develop
 
-This was developed with npm version 3.10.3 and node version 6.3.1
+This was developed with npm version 3.10.3 and node version 6.3.1.
```
summasmiff, line 18: 💅

summasmiff (approved):
Not a ton to review here, as Keytar handles it at, but it looks like you used it correctly 😄 . From what I gather, it stores the password on the user's own system, and makes calls against that - seems pretty good. Is there a reason we're still using the token rather than a more standard username + pw for this? I don't want my keychain thinking my "password" is actually a token when I go to use my keychain outside of our app.

cheshire137:
> Is there a reason we're still using the token rather than a more standard username + pw for this?

The token is preferable to requiring the user to give us their user name and password. If we had their user name and password, we could make as many tokens as we wanted, as well as do other things the user doesn't want to happen. The token limits the scope of what the app can do.

If we could do GitHub OAuth, that would be preferable to a token the user manually created for us, though. The hairiness with OAuth is the fact that this is a desktop app but is also like a web app. There are two ways we could do OAuth: implicit grant flow and authorization code flow.

Authorization code flow requires we safely store the client secret for the app. Normally in a desktop app that wouldn't be a problem, but the end user can actually get at all the contents and code of an Electron app. We can't store the client secret in the keychain with keytar, either, since the user can read those values in plain text. 

Implicit grant flow is what's normally used with web apps that can't store a client secret. The problem with using it in an Electron app is that implicit grant flow requires a URL to redirect to. We could have the app spin up a proxy server at a particular port on localhost and then have GitHub OAuth redirect to that for us to be able to grab the access token. I think eventually we'll do this. Requiring the user to create a personal access token on GitHub was just our quick-n-dirty way of starting to use the GitHub API without the hassle of setting up a proxy server so we could do proper implicit grant OAuth.

See: http://oauthlib.readthedocs.io/en/latest/oauth2/grants/implicit.html and http://stackoverflow.com/questions/16321455/what-is-the-difference-between-the-2-workflows-when-to-use-authorization-code-f

summasmiff:
Thanks, that was more than I expected. I hadn't thought of the fact that we have no place to store a client secret. Having to create a proxy server also sounds like a big hassle. It's clear that having Keytar store the token is the best situation for now, both in security and ease of use.

PR #103, opened 2016-10-15
Show a message when there are no tasks to display

### LINKED ISSUE #66: Do something with an empty task list
If a filter turns up no results, show something like "Nothing to see here" or a suggestion to better refine the search.

cheshire137:
When tasks are still loading, show a 'Loading...' message. When tasks have finished loading but there are none to display, show another message. Fixes #66.

Also changes the shortcut to toggle the developer tools to be Cmd-Option-I in macOS and Ctrl-Alt-I in Windows and Linux. This matches the shortcut in Chrome.

### DIFF
--- a/src/components/TaskList/TaskList.jsx
+++ b/src/components/TaskList/TaskList.jsx
@@ -210,13 +210,29 @@ class TaskList extends React.Component {
           </div>
         </nav>
         <div className="task-list-container">
-          <ol className="task-list">
-            {this.props.tasks.map((task, index) => {
-              const isFocused = index === this.state.selectedIndex
-              const key = `${task.storageKey}-${task.isSelected}-${isFocused}`
-              return <TaskListItem {...task} key={key} isFocused={isFocused} />
-            })}
-          </ol>
+          {this.props.tasks.length > 0 ? (
+            <ol className="task-list">
+              {this.props.tasks.map((task, index) => {
+                const isFocused = index === this.state.selectedIndex
+                const key = `${task.storageKey}-${task.isSelected}-${isFocused}`
+                return (
+                  <TaskListItem
+                    {...task}
+                    key={key}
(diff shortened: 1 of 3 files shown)


src/components/TaskList/TaskList.jsx
```diff
@@ -210,13 +210,29 @@ class TaskList extends React.Component {
...
+                return (
+                  <TaskListItem
+                    {...task}
+                    key={key}
+                    isFocused={isFocused}
+                  />
+                )
+              })}
+            </ol>
+          ) : (
```
summasmiff, line 227: Can this be made into an if/else statement instead of a ternary for readability?

src/models/AppMenu.js
```diff
@@ -97,7 +98,7 @@ class AppMenu extends EventEmitter {
       submenu: [
         {
           label: 'Developer Tools',
-          accelerator: 'CmdOrCtrl+Shift+I',
+          accelerator: `CmdOrCtrl+${this.altOrOption}+I`,
```
summasmiff, line 101: oh thank goodness, I thought this was me 😆

summasmiff (approved):
👍

src/components/TaskList/TaskList.jsx
cheshire137: I don't think it can, actually, not while staying mixed into the HTML here. I could make a function that does if/else in it and returns different chunks of HTML, then call the function here like `{this.listOrMessage()}`.

summasmiff: I think that would be preferable, for me at least. Plus, then it would be super easy to add in a little spinny icon or some other loading indicator.

cheshire137: I addressed this over in #99. 👍

PR #106, opened 2016-10-16
Fix bad app name and version, LESS compilation woes, and add app packaging

### LINKED ISSUE #33: Package executables
We need to package executables for macOS, Windows, and Linux and put them in a release. I'm thinking something like the following in our package.json:

``` json
"build-osx": "if [ -d dist/GHSnooze-darwin-x64 ]; then rm -r dist/GHSnooze-darwin-x64; fi && electron-packager src --out=dist --app-version=$npm_package_version --prune --asar --overwrite --appname=GHSnooze --platform=darwin --arch=all --ignore=test",
"build-windows": "node node_modules\\electron-packager\\cli.js src --out=dist --prune --asar --overwrite --appname=GHSnooze --platform=win32 --arch=all --ignore=test",
"build-linux": "if [ -d dist/GHSnooze-linux-ia32 ]; then rm -r dist/GHSnooze-linux-ia32; fi && if [ -d dist/GHSnooze-linux-x64 ]; then rm -r dist/GHSnooze-linux-x64; fi && electron-packager src --out=dist --app-version=$npm_package_version --prune --asar --overwrite --appname=GHSnooze --platform=linux --arch=all --ignore=test",
```

These commands don't quite work, electron-packager wasn't looking in the right place for package.json so it couldn't figure out the name or version.

### LINKED ISSUE #36: Fix imported LESS not rebuilding when it's changed

### LINKED ISSUE #58: App doesn't know its own name or version
In #32, a bug was introduced where the app thinks its name is "Electron" and thinks the version is 1.3.1, even though the package.json clearly has "GitHub Notifications Snoozer" and 0.0.1. I think this might be something to do with using electron-prebuilt-compile?

cheshire137:
This fixes #58, the app having the wrong name and version, by changing the start script to use `electron .` instead of `electron src/main.js`.

This fixes #36, the LESS not recompiling except when you change App.less, by stopping use of electron-prebuilt-compile and instead compiling LESS ourselves via less-watch-compiler.

This fixes #33 by adding scripts to package the app for Windows, Linux, and macOS.

After this lands, you should:

``` bash
rm -r node_modules
npm install
./node_modules/.bin/electron-rebuild
```

As I was trying this on Windows, I discovered #107.


src/components/App/App.less
```diff
@@ -1,19 +1,18 @@
...
 @import "../Auth/Auth.less";
+@import "../About/About.less";
 @import "../TabbedNav/TabbedNav.less";
 @import "../FilterSuggester/FilterSuggester.less";
 @import "../HiddenTaskList/HiddenTaskList.less";
 @import "../HiddenTaskListItem/HiddenTaskListItem.less";
 
 body {
-  background-color: #FFFfff;
+  background-color: #fff;
```
summasmiff, line 15: Will this be the last time this changes? 😝

package.json
```diff
@@ -47,7 +47,7 @@
     "type": "git",
     "url": "git+https://github.com/cheshire137/gh-notifications-snoozer.git"
   },
-  "author": "probablycorey and cheshire137",
+  "author": "probablycorey, cheshire137, summasmiff",
```
summasmiff, line 58: ❤️

package.json
```diff
@@ -31,17 +32,19 @@
...
+    "less-watch-compiler": "^1.2.3",
     "mocha": "^2.5.3",
     "react-addons-test-utils": "^15.2.1",
     "redux-devtools": "^3.3.1",
     "xvfb-maybe": "^0.1.3"
   },
   "scripts": {
     "postinstall": "node postinstall.js",
-    "start": "electron src/main.js",
+    "start": "./node_modules/.bin/concurrently \"npm run build-css\" \"./node_modules/.bin/electron src/main.js\"",
```
summasmiff: 👍

summasmiff (approved):
Looks great! Thanks for adding me :)

PR #109, opened 2016-10-22
Add keyboard shortcuts for archive, snooze, ignore

### LINKED ISSUE #50: Add keyboard shortcuts for ignore/archive/snooze
Akin to #39, it would be good to have assigned key combinations for ignoring, archiving, and snoozing selected tasks. Should probably take inspiration from Gmail or Apple Mail for similar behaviors:

screen shot 2016-09-04 at 5 32 14 pm

### LINKED ISSUE #101: Add menu with options for selected tasks
When I have one or more tasks selected and thus the archive/ignore/snooze buttons are enabled, it would be nice to see those same options repeated in a menu, perhaps a Task menu. Likewise when I'm viewing the hidden tasks list, it would be good to have a Restore option in a Task menu. These menu options would do the same thing as clicking the button.

You might knock #50 out at the same time by setting an `accelerator` for each menu option.

src/models/AppMenu.js is where the menu options should be added.

src/components/App/App.jsx is where the archive/ignore/snooze/restore events should be listened for.

The menu options should be grayed out when no tasks are selected, or when the action is not applicable to the selected task(s). For example, 'Restore' should be grayed out when a visible task is selected, and 'Archive' should be grayed out when a hidden task is selected. New functions like `setIsAuthenticated` can be added to AppMenu.js for enabling/disabling each new menu item. These can be called in App.jsx as the selected tasks change. TaskList and HiddenTaskList may need to be given new handlers for notifying App when the selected tasks change, e.g., `onSelectionChange`.

cheshire137:
Fixes #50 and fixes #101.

This adds a Task menu with options Archive, Snooze, Ignore, and Restore. Those options become enabled as necessary based on whether hidden or visible tasks are being viewed and whether or not there are any tasks selected.

To do this, the concept of 'focusing' a task to highlight which one you're looking at is now part of Redux storage instead of regular React per-component state. Also when you use the arrow keys to focus a task, it also changes task selection.

Having task focus as part of persistent Redux state may help users to keep track of where they left off.

I'm open to feedback as to the shortcuts being used. It's Ctrl+Option+letter right now for archive/snooze/ignore/restore. Maybe it should just be the letter only, e.g., A for archive.

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -87,6 +87,15 @@ class App extends React.Component {
         this.manageFilters()
       }
     })
+    this.appMenu.on('archive', () => {
+      this.props.dispatch({ type: 'TASKS_ARCHIVE' })
+    })
+    this.appMenu.on('ignore', () => {
+      this.props.dispatch({ type: 'TASKS_IGNORE' })
+    })
+    this.appMenu.on('snooze', () => {
+      this.props.dispatch({ type: 'TASKS_SNOOZE' })
+    })
   }
 
   getViewContents() {
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -135,7 +135,7 @@ class App extends React.Component {
           loadPrevPage={loadPrevPage}
           currentPage={currentPage}
           loading={this.state.loadingTasks}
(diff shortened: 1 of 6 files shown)


cheshire137:
Closing this in favor of #116. 🎩 @probablycorey.

PR #111, opened 2016-11-16
Changes to tasks

probablycorey:
This does a few things:

- It fetches the minimum number of results
  - When a filter is created, applies the "is:open" filter to grab everything
  - From then on it adds the "updated:>XYZ" to only grab tasks that have been updated
  - This allows us to remove pagination from the app.

- Only comment and state changes will mark a task as updated. An issue won't be unarchived automatically when the body or a label is changed.

There are some other minor things that I'll leave comments about below.

cc @cheshire137 @summasmiff


src/components/TaskList/TaskList.jsx
```diff
@@ -85,12 +85,15 @@ class TaskList extends React.Component {
...
-      .then(result => {
-        const tasks = result.tasks
-        const filter = this.props.activeFilter
-        this.props.dispatch({ type: 'TASKS_UPDATE', filter, tasks })
-      })
+    github.getTasks(this.props.activeFilter).then(result => {
+      const tasks = result.tasks
+
+      // GitHub's api doesn't like when the output includes milliseconds
+      const updatedAt = (new Date()).toISOString().replace(/\.\d{3}Z/, 'Z')
```
probablycorey, line 92: This makes me barf, but I couldn't find a workaround. Javascript's toISOString retruns a screwy format.

src/components/TaskList/TaskList.jsx
```diff
@@ -129,10 +132,14 @@ class TaskList extends React.Component {
   }
 
   taskListOrMessage() {
+    const sortedTasks = this.props.tasks.sort((a, b) => {
+      return Date.parse(b.updatedAt) - Date.parse(a.updatedAt)
```
probablycorey, line 136: Always sort tasks by `updatedAt` for now.

src/models/GitHub.js
```diff
@@ -54,22 +57,27 @@ class GitHub extends Fetcher {
...
-    return this.get(url).then(result => {
-      const { json, nextUrl } = result
-      return { tasks: json.items.map(d => getTask(d)), nextUrl,
-               currentUrl: url }
+  getTasksFromUrl(url, items = []) {
+    return this.get(url).then(({ json, headers }) => {
+      const allItems = items.concat(json.items)
+      const nextUrl = this.getNextUrl(headers)
+      if (nextUrl) {
+        return this.getTasksFromUrl(nextUrl, allItems)
```
probablycorey, line 77: Fetches all tasks, 100 tasks per page. But because subsequent fetches will use the `updated:>${filter.updatedAt}` filter, I don't think it will auto-paginate very often.

src/reducers/TasksReducer.js
```diff
@@ -1,20 +1,37 @@
...
   // Add the existing existingTasks
   existingTasks.forEach(task => (tasksByKey[task.storageKey] = task))
 
   // Update existingTasks with new values and add new existingTasks
-  tasks.forEach((newTask) => {
-    const oldTask = tasksByKey[newTask.storageKey] || {}
-    const filterQueries = oldTask.filterQueries || []
-    if (!filterQueries.includes(filter.query)) {
-      filterQueries.push(filter.query)
+  tasks.forEach((updatedTask) => {
```
probablycorey: This method is getting too complicated and could use a refactor.

test/fixtures.js
```diff
@@ -242,4 +272,10 @@ const notification = {
...
 }
 
-module.exports = { task, pullRequest, issue, notification }
+module.exports = {
+  task: Object.freeze(task),
+  anotherTask: Object.freeze(anotherTask),
+  pullRequest: Object.freeze(pullRequest),
+  issue: Object.freeze(issue),
+  notification: Object.freeze(notification),
+}
```
probablycorey, line 281: I wanted to try and prevent myself from accidentally changing filters. Freeze isn't enough though, we need something like deepFreeze.

test/reducers/TasksReducerTest.js
```diff
@@ -1,4 +1,5 @@
-const assert = require('assert')
+const { assert } = require('chai')
```
probablycorey, line 1: Switched to chai because it seems like the standard mocha assert framework and it gives us some better assertions.

cheshire137:
> to only grab tasks that have been updated

Updated since when?

package.json
```diff
@@ -44,7 +46,7 @@
     "start": "concurrently \"npm run watch-css\" \"electron .\"",
     "style": "eslint -c .eslintrc *.js scripts/* src/*.js src/**/*.jsx test/**/*.js",
     "unit-tests": "xvfb-maybe electron-mocha --renderer --recursive --compilers js:babel-register test",
-    "test": "npm run-script style && NODE_ENV=test npm run-script unit-tests",
+    "test": "npm run-script style --silent && NODE_ENV=test npm run-script unit-tests --silent",
```
cheshire137, line 49: What does the silent option do?

src/components/TaskList/TaskList.jsx
cheshire137, line 92: Haha considering your comment, I bet there's a story here about how you discovered this particular limitation.

probablycorey:
> Updated since when?

Since the last time that filter query was fetched by the app.

src/components/TaskList/TaskList.jsx
cheshire137: So the `sort:` stuff we should remove from the search autocompletion?

src/components/TaskListItem/TaskListItem.jsx
```diff
@@ -67,10 +68,21 @@ class TaskListItem extends React.Component {
     return listItemClasses.join(' ')
   }
 
+  changes(changelog) {
+    if (Object.keys(changelog || {}).length !== 0) {
```
cheshire137: I'd say `> 0` reads better.

src/models/GitHub.js
```diff
@@ -54,22 +57,27 @@ class GitHub extends Fetcher {
...
   }
 
   // https://developer.github.com/v3/search/#search-issues
-  getTasks(query) {
-    const params = `?q=${encodeURIComponent(query)}`
-    const url = `${Config.githubApiUrl}/search/issues${params}`
+  getTasks(filter) {
+    const extraParams = filter.updatedAt ? ` updated:>${filter.updatedAt}` : ' is:open'
+    const params = `?q=${encodeURIComponent(filter.query + extraParams)}`
+    const url = `${Config.githubApiUrl}/search/issues${params}&per_page=100`
```
cheshire137, line 68: So we'll only see the most recently updated 100 issues/PRs? I think this makes sense if the goal is to continuously be archiving/ignoring/snoozing tasks.

src/models/GitHub.js
```diff
@@ -54,22 +57,27 @@ class GitHub extends Fetcher {
...
     return this.getTasksFromUrl(url)
   }
 
-  getTasksFromUrl(url) {
-    return this.get(url).then(result => {
-      const { json, nextUrl } = result
-      return { tasks: json.items.map(d => getTask(d)), nextUrl,
-               currentUrl: url }
+  getTasksFromUrl(url, items = []) {
+    return this.get(url).then(({ json, headers }) => {
```
cheshire137, line 73: Nice, I like extracting the properties here rather than my `const { json, nextUrl } = result`. 👍

src/reducers/TasksReducer.js
cheshire137: You mentioned it yesterday, but I personally like the more terse style of not having the parentheses around a single parameter, and only adding them when we have to when we have multiple parameters.

test/reducers/TasksReducerTest.js
```diff
@@ -109,26 +65,35 @@ describe('Tasks reducer', () => {
...
-      assert(!actual[1].isSelected, 'task should not be selected')
-      assert.equal('string', typeof actual[1].archivedAt,
-                   'archivedAt should be a time string')
+      const archivedTask = store.getState().tasks.find(t => t.archivedAt)
+
+      assert(!archivedTask.isSelected, 'archived tasks should not be selected')
+      assert.isNotNaN(Date.parse(archivedTask.archivedAt), 'archivedAt should be a string date')
+    })
+
+    it('clear changelog when task is archived', () => {
```
cheshire137: Maybe "clears" so it reads better? "it clears changelog"

src/models/GitHub.js
probablycorey: We'll see them all, but we get 100 results per page. For example, one of my filters is `involves:cheshire137` and I get back 4 pages and 395 issues and PRs.

package.json
probablycorey: It gets rid of this annoying npm error

img

src/components/TaskList/TaskList.jsx
probablycorey: Yeah, that makes sense.

src/reducers/TasksReducer.js
probablycorey: I added the `"arrow-parens": ["error", "as-needed"]` eslint rule to enforce your preference on everyone!

PR #115, opened 2016-11-18
Update tasks automatically

probablycorey:
This solves three problems:

1. Every 60 seconds it will iterate through all filters and fetch the updated tasks
2. When you launch the app it will iterate through all filters and fetch the updated tasks
3. When you update a filter it will fetch the updated tasks for that filter

To do this I pulled the some dispatches into a class called `HelperActions`. This is [common in Redux](http://redux.js.org/docs/basics/Actions.html). What's not common is passing in the dispatcher to the actions. But these actions require multiple dispatches, so I couldn't think of a good way around it.

@cheshire137

### DIFF
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -119,9 +119,7 @@ class App extends React.Component {
 
     let promise = Promise.resolve()
     this.props.filters.forEach(filter => {
-      promise = promise.then(() => {
-        return HelperActions.updateTasks(this.props.dispatch, filter)
-      })
+      promise = promise.then(() => HelperActions.updateTasks(this.props.dispatch, filter))
     })
 
     return promise.then(() => {
@@ -219,7 +217,7 @@ App.propTypes = {
 const mapStateToProps = state => {
   const props = {
     filters: state.filters,
-    activeFilter: state.filters.find(filter => filter.selected)
+    activeFilter: state.filters.find(filter => filter.selected),
   }
   return props
 }
--- a/src/components/App/App.jsx
+++ b/src/components/App/App.jsx
@@ -4,6 +4,7 @@ const { ipcRenderer } = require('electron')
 const GitHub = require('../../models/GitHub')
 const AppMenu = require('../../models/AppMenu')
 const GitHubAuth = require('../../models/GitHubAuth')
+const HelperActions = require('../../models/HelperActions')
 const TabbedNav = require('../TabbedNav')
 const DefaultFilters = require('../../models/DefaultFilters')
 const TaskList = require('../TaskList')
@@ -29,6 +30,8 @@ class App extends React.Component {
   }
 
   componentDidMount() {
+    this.updateTasksInBackground()
+
     ipcRenderer.send('title', 'Tasks')
     this.setupAppMenu()
     if (GitHubAuth.isAuthenticated()) {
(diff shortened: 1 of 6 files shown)


src/index.jsx
```diff
@@ -10,6 +10,23 @@ import { remote } from 'electron'
 import { persistStore, autoRehydrate } from 'redux-persist'
 import { AsyncNodeStorage } from 'redux-persist-node-storage'
 
+// This hack makes my blood boil. But without it the intial state is
+// used instead of using the persisted state from AsyncNodeStorage. I'd prefer
+// to use something other than setTimeout to make this work, but it works for
+// now.
+const hackToLoadEverything = store => {
```
probablycorey, line 17: I'm not happy with this approach, but after a bit of research I failed to find a better one. For not I think it's fine.

src/components/App/App.jsx
```diff
@@ -111,6 +114,20 @@ class App extends React.Component {
     }
   }
 
+  updateTasksInBackground() {
+    if (!GitHubAuth.isAuthenticated()) return
+
+    let promise = Promise.resolve()
+    this.props.filters.forEach(filter => {
+      promise = promise.then(() => HelperActions.updateTasks(this.props.dispatch, filter))
```
cheshire137, line 122: Interesting, so this is chaining them all off of each other in `then` callbacks? What I've done before is to use [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) on an array of promises. So you could do `this.props.filters.map(filter => HelperActions.updateTasks(this.props.dispatch, filter))` if updateTasks returns a promise, and pass the resulting array to Promise.all.

src/components/TaskList/TaskList.jsx
```diff
@@ -133,7 +124,9 @@ class TaskList extends React.Component {
 
   taskListOrMessage() {
     const sortedTasks = this.props.tasks.sort((a, b) => {
-      return Date.parse(b.updatedAt) - Date.parse(a.updatedAt)
+      const dateA = Date.parse(a.updatedAt)
+      const dateB = Date.parse(b.updatedAt)
+      return dateB - dateA
```
cheshire137: Why expand this to three lines? I thought the one line was pretty readable.

src/index.jsx
```diff
@@ -10,6 +10,23 @@ import { remote } from 'electron'
 import { persistStore, autoRehydrate } from 'redux-persist'
 import { AsyncNodeStorage } from 'redux-persist-node-storage'
 
+// This hack makes my blood boil. But without it the intial state is
```
cheshire137: `intial` => `initial`

src/index.jsx
```diff
@@ -10,6 +10,23 @@ import { remote } from 'electron'
...
+  const render = () => {
+    ReactDom.render(
+      <Provider store={store}>
+        <App />
+      </Provider>,
+      document.getElementById('content')
+    )
+  }
+
+  setTimeout(render, 10)
```
cheshire137, line 27: Haha wow.

cheshire137 (approved):

src/components/App/App.jsx
probablycorey: I wanted to use Promise.all, but that was running all the fetches in parallel. It wasn't usually a problem, but at one point it caused an API error because I went over the rate limit. So I ended up with this approach so it runs in serial. 

I couldn't find a serial equivalent to Promise.all so I went with this.

src/components/TaskList/TaskList.jsx
probablycorey: I agree. In a moment of weakness I did this to appease the `arrow-body-style` linting warning. I tried to find a way to fix this in the .eslintrc file, but after 10 minutes I gave up.

I'll put it back to one line.

src/index.jsx
probablycorey: setTimeout... the nuclear option.

PR #116, opened 2016-11-19
Add keyboard shortcuts and remove selected tasks from state 

### LINKED ISSUE #50: Add keyboard shortcuts for ignore/archive/snooze

probablycorey:
This PR got away from me a little bit, I can split it into two PRs if it is too much code to review. 

I :heart: the ideas from https://github.com/cheshire137/gh-notifications-snoozer/pull/109/files, but it has become too hard to merge. So I added those keyboard shortcuts to this PR.

This PR also:

1. Based on some feedback from @cheshire137, this removes the ability to select multiple tasks. After trying this out I feel like the app is much easier to use!

2. It gets rid of the hidden task components and moves that logic into TaskList. I did this because they started to feel so similar. The downside to this is TaskList has become too big, but I have plans to fix that in a follow-up PR.

Closes #50


probablycorey:
FYI: There are still a few bugs I'm working out here.

probablycorey:
@cheshire137 the bugs are worked out and I think this is good to go. I started to implement menu based keyboard shortcuts too, but it started to become a lot of extra code. So I think that could be done in a separate PR.

src/components/HiddenTaskList/HiddenTaskList.jsx
```diff
@@ -1,83 +0,0 @@
-const React = require('react')
```
cheshire137, line 1: I ❤️ that you were able to absorb the hidden task list stuff into TaskList. I wanted that when I first implemented HiddenTaskList but hit problems that I've forgotten now when I first tried it.

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
-    this.state = { selectedIndex: null }
-    this.onKeyUp = this.onKeyUp.bind(this)
-    this.onKeyDown = this.onKeyDown.bind(this)
+    this.state = { focusedIndex: 0, showHidden: false }
   }
 
   componentDidMount() {
-    document.addEventListener('keyup', this.onKeyUp)
-    document.addEventListener('keydown', this.onKeyDown)
+    this.keyDownEventListener = event => this.onKeyDown(event)
```
cheshire137, line 16: Why use the style of creating a new property on the class rather than `this.onKeyDown = this.onKeyDown.bind(this)`? Seems neater to me to keep just the one property and rebind it.

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
   }
 
-  onSnoozeClick(event) {
-    event.currentTarget.blur() // defocus button
-    this.props.dispatch({ type: 'TASKS_SNOOZE' })
+  onKeyDown(event) {
+    const up = ['ArrowUp', 'k']
+    const down = ['ArrowDown', 'j']
+    const open = ['Enter', 'o']
+    const select = [' ', 'x']
```
cheshire137, line 28: Is `x` the Gmail shortcut?

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
-    document.removeEventListener('keydown', this.onKeyDown)
+    document.removeEventListener('keydown', this.keyDownEventListener)
   }
 
-  onSnoozeClick(event) {
-    event.currentTarget.blur() // defocus button
-    this.props.dispatch({ type: 'TASKS_SNOOZE' })
+  onKeyDown(event) {
+    const up = ['ArrowUp', 'k']
+    const down = ['ArrowDown', 'j']
```
cheshire137: Those vim bindings! 😀

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
+
+    let preventDefault = true
+
+    if (up.includes(event.key)) {
+      this.focusPreviousTask()
+    } else if (down.includes(event.key)) {
+      this.focusNextTask()
+    } else if (escape.includes(event.key)) {
+      this.setState({ focusedIndex: 0 })
+    } else if (open.includes(event.key)) {
```
cheshire137, line 42: You kind of mix action with particular keys with your variable names, e.g., compare `up` with `open`. Maybe make each of your vars be named after the action, then the conditions will read like this one, `open.includes(event.key)`. So `up` => `previous`, `down` => `next`, `escape` => `defocus`.

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
+    if (preventDefault) event.preventDefault()
   }
 
-  onArchiveClick(event) {
-    event.currentTarget.blur() // defocus button
-    this.props.dispatch({ type: 'TASKS_ARCHIVE' })
+  tasks() {
+    return this.props.allTasks.filter(task => {
+      const matchesFilter = (task.filterQueries || []).includes(this.props.activeFilter.query)
+      if (!matchesFilter) return false
```
cheshire137: Oh man, these one-liner inline if statements make me a bit nervous. 😅 We're not in CoffeeScript land!

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
+  focusNextTask() {
+    const lastIndex = this.tasks().length - 1
+    let focusedIndex = this.state.focusedIndex + 1
+    if (focusedIndex > lastIndex) focusedIndex = 0
+    this.setState({ focusedIndex })
+  }
+
+  focusPreviousTask() {
+    const lastIndex = this.tasks().length - 1
+    let focusedIndex = this.state.focusedIndex + -1
```
cheshire137: `+ -1`? Haha.

src/components/TaskList/TaskList.jsx
```diff
@@ -9,68 +9,103 @@ const TaskVisibility = require('../../models/TaskVisibility')
...
+      preventDefault = false
+    }
+
+    if (preventDefault) event.preventDefault()
   }
 
-  onArchiveClick(event) {
-    event.currentTarget.blur() // defocus button
-    this.props.dispatch({ type: 'TASKS_ARCHIVE' })
+  tasks() {
```
cheshire137, line 61: Is this a method as opposed to filtering and passing only the applicable tasks as part of the properties list to the component so that we're able to change which tasks are visible on the fly? You couldn't re-render the whole component with new properties when hidden/visible changes? I don't even know if that's possible, just wondering.

src/components/TaskList/TaskList.jsx
```diff
@@ -87,63 +122,82 @@ class TaskList extends React.Component {
...
-    const oldIndex = this.state.selectedIndex
-    const lastIndex = this.props.tasks.length - 1
-    let newIndex = typeof oldIndex === 'number' ? oldIndex - 1 : lastIndex
-    if (newIndex < 0) {
-      newIndex = lastIndex
+  navRight() {
+    if (this.state.showHidden) {
+      return (
+        <div className="nav-right">
+          <div className="nav-right">
```
cheshire137: Why the nested `nav-right` divs?

src/components/TaskList/TaskList.jsx
```diff
@@ -87,63 +122,82 @@ class TaskList extends React.Component {
...
-            const isFocused = index === this.state.selectedIndex
-            const key = `${task.storageKey}-${task.isSelected}-${isFocused}`
-            return <TaskListItem task={task} key={key} isFocused={isFocused} />
+          {this.tasks().map((task, index) => {
+            const isFocused = (index === this.state.focusedIndex)
+            return (
+              <TaskListItem
+                task={task}
+                key={task.storageKey}
+                isFocused={isFocused}
```
cheshire137, line 196: If you wanted, you could drop the curly braces on the `map` block, drop the `isFocused` variable in favor of doing the logic inline here, and drop the `return`:

```javascript
{this.tasks().map((task, index) => (
  <TaskListItem
```

src/components/TaskList/TaskList.jsx
```diff
@@ -168,47 +222,17 @@ class TaskList extends React.Component {
...
-              <button
-                type="button"
-                className="control button is-link"
-                onClick={e => this.onIgnoreClick(e)}
-                title="Ignore selected"
-                disabled={isIgnoreDisabled}
-              >❌ Ignore</button>
-            </span>
-          </div>
+          {this.navRight()}
```
cheshire137, line 231: I like that you pulled this into a method to make this chunk easier to read.

src/components/TaskList/TaskList.jsx
```diff
@@ -219,7 +243,9 @@ class TaskList extends React.Component {
           </div>
         </nav>
         <div className="task-list-container">
-          {this.taskListOrMessage()}
+          <div className={this.state.showHidden && 'viewing-hidden'}>
```
cheshire137, line 252: Does this result in the class being nothing if showHidden is false, and the class being 'viewing-hidden' when showHidden is true?

src/reducers/TasksReducer.js
```diff
@@ -70,78 +44,51 @@ function clearChangelog(existingTasks, { task }) {
...
-        snoozedAt: null,
-        isSelected: false,
-        ignore: false,
-        changelog: {},
-      })
-    }
-    return task
-  })
-  return updatedTasks
+function ignoreTasks(existingTasks, { task: taskToUpdate }) {
```
cheshire137: It seems odd that the method name is plural but it only ever works on a single task.

test/components/TaskListTest.js
```diff
@@ -43,68 +43,33 @@ describe('TaskList', () => {
...
-    store.dispatch({ type: 'TASKS_IGNORE' })
+    store.dispatch({ type: 'TASKS_IGNORE', task: { storageKey: 'pull-163031382' } })
     assert.equal(0, renderedDOM().querySelectorAll('#pull-163031382').length)
-
-    // Reset state
-    store.dispatch({
-      type: 'TASKS_SELECT', task: { storageKey: 'pull-163031382' },
-    })
-    store.dispatch({ type: 'TASKS_RESTORE' })
+    store.dispatch({ type: 'TASKS_RESTORE', task: { storageKey: 'pull-163031382' } })
```
cheshire137, line 48: I like that we're using the action part now so we don't have to rely on having previously sent a different action to affect a task. It's nice to have it all in one command.

test/components/TaskListTest.js
```diff
@@ -163,23 +114,6 @@ describe('TaskList', () => {
       const task = store.getState().tasks[0]
       assert.equal('string', typeof task.archivedAt)
     })
-
-    it('shows tasks again if `updatedAt` is greater than `archivedAt`', () => {
```
cheshire137, line 167: Is this no longer the case?

test/components/TaskListTest.js
```diff
@@ -136,22 +101,8 @@ describe('TaskList', () => {
...
-
       TestUtils.Simulate.click(renderedDOM().querySelector('#archive-button'))
-      archiveTime = new Date()
-    })
-
-    after(() => {
-      store.dispatch({
-        type: 'TASKS_SELECT', task: { storageKey: 'pull-163031382' },
-      })
-      store.dispatch({ type: 'TASKS_RESTORE' })
```
cheshire137, line 154: The before block is still clicking the archive button. Do we not need to restore a task in the after block?

cheshire137 (approved):
Love that we're going to get key shortcuts for archive/ignore/snooze, that HiddenTaskList is going away while the functionality is still there, and that you dropped task checkboxes! ❤️

src/components/TaskList/TaskList.jsx
probablycorey: I switched to this because it felt weird to reassign an instance class from within the instance. I'd prefer if we could just do 

```js
 document.addEventListener('keydown', this.onKeyDown)

// and

 document.removeEventListener('keydown', this.onKeyDown)
```

But Javascript is evil and likes to make things difficult for us :sad:

src/components/TaskList/TaskList.jsx
probablycorey: I'll never give them up!

src/components/TaskList/TaskList.jsx
probablycorey: Yeah. It probably makes sense to add this to the config file, but for now I'm just hard coding them.

src/components/TaskList/TaskList.jsx
probablycorey: > Is this a method as opposed to filtering and passing only the applicable tasks as part of the properties list to the component so that we're able to change which tasks are visible on the fly? 

Yeah, when the components hidden state changes, this makes sure the right tasks are rendered.

> You couldn't re-render the whole component with new properties when hidden/visible changes?

So you're saying, if the hidden state changes, the JSX that renders the TaskList would be passed the updated list of tasks? We could do that, but right now the hidden/unhidden button is on the TaskList so that is where all the magic has to happen.

src/components/TaskList/TaskList.jsx
probablycorey: Yeah, I usually avoid those, but Ruby has me wanting them back. If you make a linter rule I'll promise to always follow it!

src/components/TaskList/TaskList.jsx
probablycorey: 😮

src/components/TaskList/TaskList.jsx
probablycorey: Copy-Pasta error. Thanks for finding it!

src/components/TaskList/TaskList.jsx
probablycorey: I dig it!

src/components/TaskList/TaskList.jsx
probablycorey: I was very close to making a new `TaskNavBar` component.

src/components/TaskList/TaskList.jsx
probablycorey: Yeah, I learned that is a JSX thing. If the className is falsey, it doesn't add any class.

src/reducers/TasksReducer.js
probablycorey: That is weird...

test/components/TaskListTest.js
probablycorey: The store is thrown out and recreated after each test, so we don't have to clean anything up!

test/components/TaskListTest.js
probablycorey: It should totally be the case. It looks like I tried to move it but never did :sad:. I'll add it back in.

PR #119, opened 2016-11-21
Don't always require electron-devtools-installer

### LINKED ISSUE #117: Can't find electron-devtools-installer in packaged app
When I run `npm run build-macos` and run the resulting GHSnooze.app, this error appears:

screen shot 2016-11-21 at 10 14 37 am

I think maybe we're always initializing the devtools stuff and we shouldn't. Conversely, we could move it from devDependencies to dependencies in package.json, but it seems like the kind of thing we don't want in the final product.

/cc @probablycorey

cheshire137:
The electron-devtools-installer package can be useful when developing, but when you just want to run the app, you shouldn't need it installed. Fixes #117.

/cc @probablycorey @shayfrendt

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,7 +17,8 @@
     "no-console": 0,
     "react/prefer-stateless-function": 0,
     "import/no-unresolved": 0,
-    "arrow-parens": ["error", "as-needed"]
+    "arrow-parens": ["error", "as-needed"],
+    "global-require": 0
   },
   "ecmaFeatures": {
     "jsx": true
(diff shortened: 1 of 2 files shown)


shayfrendt:
The app is building and opening successfully for me now, but no GUI shows up when I do open the app via `open dist/GHSnooze-darwin-x64/GHSnooze.app/`. Is there something else I need to do or configure?

shayfrendt:
Hmm closing and re-opening it shows a blank window now:

screen shot 2016-11-21 at 6 47 37 pm

PR #124, opened 2016-12-22
Start using GraphQL

probablycorey:
This moves Snoozer off the REST API and to GraphQL. There are a few advantages with this:

- We can now know if a PR is merged!
- We can get all the new comments for an Issue or PR. This will allow us to keep an issue/pr archived if you were the one who commented on it.
- Less Code and GraphQL is the future!

There are two hiccups. You have to visit https://github.com/prerelease/agreement for the GraphQL API to work! And I need to get the tests to pass.

cc @cheshire137


src/models/GitHub.js
```diff
@@ -1,157 +1,158 @@
...
+    return fetch(url, options).then(response => {
+      if (!response.ok) {
+        return response.json().then(json => {
+          throw new Error(`${response.status}: ${JSON.stringify(json)}`)
+        })
+      }
+
+      return response.json().then(result => {
+        if (result.errors) {
+          throw new Error(`GraphQL Error: ${result.errors.map(e => e.message)}`)
```
cheshire137: Maybe a `.join(', ')` after the `map` call.

src/models/GitHub.js
```diff
@@ -1,157 +1,158 @@
...
   }
 
   getHeaders() {
     if (!this.token) {
       this.token = GitHubAuth.getToken()
     }
     return {
-      Accept: 'application/vnd.github.v3+json',
-      Authorization: `token ${this.token}`,
+      Authorization: `bearer ${this.token}`,
```
cheshire137, line 72: Interesting. Do you know why we change from `token` in the header to `bearer`?

src/models/GitHub.js
```diff
@@ -1,157 +1,158 @@
...
+      url: node.url,
+      number: node.number,
+      repository: `${node.repository.owner.login}/${node.repository.name}`,
+      repositoryOwner: node.repository.owner.login,
+      repositoryOwnerUrl: node.repository.owner.url,
+      repositoryOwnerAvatar: node.repository.owner.avatarURL,
+      user: node.author.login,
+      userUrl: node.author.url,
+      userAvatar: node.author.avatarURL,
+      comments: node.comments ? node.comments.totalCount : 0,
```
cheshire137, line 98: Is this preserving the existing behavior where if you're the one who comments on a task, it will resurface in snoozer? I think that's fine, I'd rather introduce GraphQL in this branch and make the fix for not notifying yourself about your own comments in a follow-up branch.

src/models/GitHub.js
```diff
@@ -1,157 +1,158 @@
...
-    let nextUrl
-    urlsAndRels.forEach(str => {
-      const urlAndRel = str.trim().split('; ')
-      if (urlAndRel[1] === 'rel="next"') {
-        const urlInBrackets = urlAndRel[0]
-        nextUrl = urlInBrackets.slice(1, urlInBrackets.length - 1)
-        return
+  taskQuery() {
+    return `query($search: String!, $endCursor: String) {
+      search(first: 30, query: $search, after: $endCursor, type: ISSUE) {
```
cheshire137: I wouldn't mind seeing the 30 pulled out into a named const, something like `const TASKS_PER_PAGE = 30`.

src/models/GitHub.js
```diff
@@ -1,157 +1,158 @@
...
+        pageInfo {
+         endCursor,
+          hasNextPage
+        }
+        edges {
+          node {
+            ... on Issueish {
```
cheshire137, line 112: Was there documentation you were following about the types available? If so, maybe link to it in a comment here but also in the readme.

cheshire137 (commented):
I'll re-review when tests are passing. I would add a link to https://github.com/prerelease/agreement in the readme since users will need to sign it, not just those developing the app.

src/models/GitHub.js
probablycorey: I saw they used bearer in the graphql docs, so I updated it to match that. I'm not sure if it matters though.

src/models/GitHub.js
probablycorey: > Is this preserving the existing behavior where if you're the one who comments on a task, it will resurface in snoozer?

Yeap! I tried to keep that the same and fix it in a follow-up!

